PR #2, opened 2019-07-04
refactor bitops2, add endians2

arnetheduck:


std_shims/support/endians2.nim
```diff
@@ -0,0 +1,181 @@
...
+    array[sizeof(x), byte] {.inline.} =
+  ## Convert a native endian integer to a little endian byte sequence
+  toBytes(x, littleEndian)
+
+func toBytesBE*(x: uint8|uint16|uint32|uint64):
+    array[sizeof(x), byte] {.inline.} =
+  ## Convert a native endian integer to a native endian byte sequence
+  toBytes(x, bigEndian)
+
+func fromBytes*(
```
cheatfate: I think `openarray[byte]` as `x` argument allow much more handy API, but of course it will require bounds check which can be done without errors/exceptions. So if there empty `openarray[byte]` result value is `0`.

cheatfate:
I'm just curious is bitops2.nim and endians2.nim performs only compile-time tests or runtime tests too?

arnetheduck:
> I'm just curious is bitops2.nim and endians2.nim performs only compile-time tests or runtime tests too?

both: https://github.com/status-im/nim-std-shims/pull/2/files#diff-85b7474486b0a6449b657bb71b1411d9R54

stew.nimble
```diff
@@ -10,6 +10,12 @@ skipDirs      = @["tests"]
 requires "nim >= 0.19.0"
 
 task test, "Run all tests":
+<<<<<<< HEAD
```
tersec: Merge artifacts?

arnetheduck: huh, how did that happen? pushing..

PR #7, opened 2019-09-19
WIP: fix error with ref types

Tormund:


Tormund:
Faced with issue in json_serialization:
```
import json_serialization

type
    DummyRef* = ref object of RootObj
        a: float
var dr = new(DummyRef)
dr.a = 100225.00
echo "resref:\n\t", dr.toJson 
```
nim-stew\stew\shims\macros.nim(43, 28) Error: index 2 not in 0 .. -1

PR #8, opened 2019-09-20
ptrops: add

arnetheduck:
This adds ptr ops to `stew` - it has previously been discussed in relation to `nim-serialization`, `ssz` and std library.. 

Names are long on purpose so that one can look for them in code, and follow naming conventions drawn from C++ and rust - in particular, `shift` is typically associated with bit shifting.

### DIFF
--- a/stew/ptrops.nim
+++ b/stew/ptrops.nim
@@ -35,9 +35,8 @@ template offset*[T](p: ptr T, count: int): ptr T =
 
   # We turn off checking here - too large counts is UB
   {.checks: off.}
-  mixin offset
   let bytes = count * sizeof(T)
-  cast[ptr T](cast[pointer](p).offset(bytes))
+  cast[ptr T](offset(cast[pointer](p), bytes))
 
 template distance*(a, b: pointer): int =
   ## Number of bytes between a and b - undefined behavior when difference
@@ -50,5 +49,4 @@ template distance*[T](a, b: ptr T): int =
   # Number of elements between a and b - undefined behavior when difference
   # exceeds what can be represented in an int
   {.checks: off.}
-  mixin toMemAddress, distance
-  cast[pointer](a).distance(cast[pointer](b)) div sizeof(T)
+  distance(cast[pointer](a), cast[pointer](b)) div sizeof(T)
--- a/stew/ptrops.nim
+++ b/stew/ptrops.nim
@@ -11,60 +11,44 @@
 # been offset is unsafe and many of the operations herein have undefined
 # and platform-specific behavior in corner cases.
 
-# Due to poor codegen and subsequent lack of inlining, many of these operations
+# Due to poor codegen and subsequent lack of inlining, these operations
 # are templates even where they could be func.
 
-# ByteAddress in std lib is signed - this leads to issues with overflow checking
-# when address is on boundary.
-type
-  MemAddress* = distinct uint
-
-template toMemAddress*(p: pointer): MemAddress = cast[MemAddress](p)
-template toPointer*(p: MemAddress): pointer = cast[pointer](p)
-template toPtr*(p: MemAddress, T: type): ptr T = cast[ptr T](p)
-
-template offset*(p: MemAddress, bytes: int): MemAddress =
-  ## Offset a memory address by a number of bytes. Behavior is undefined on
-  ## overflow.
-  # Actual behavior is wrapping, but this may be revised in the future to enable
-  # better optimizations
-  {.checks: off.}
-  MemAddress(uint(p) + cast[uint](bytes))
+# Note that `ByteAddress` in stdlib is implemented as a signed integer and
+# might lead to overflow on arithmetic - avoid
 
 template offset*(p: pointer, bytes: int): pointer =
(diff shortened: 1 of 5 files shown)


stew/ptrops.nim
```diff
@@ -0,0 +1,71 @@
...
+template toPointer*(p: MemAddress): pointer = cast[pointer](p)
+template toPtr*(p: MemAddress, T: type): ptr T = cast[ptr T](p)
+
+template offset*(p: MemAddress, bytes: int): MemAddress =
+  ## Offset a memory address by a number of bytes. Behavior is undefined on
+  ## overflow.
+  # Actual behavior is wrapping, but this may be revised in the future to enable
+  # better optimizations
+  {.checks: off.}
+  mixin offset
```
jangko: I don't get it, why a mixin here?

stew/ptrops.nim
```diff
@@ -0,0 +1,71 @@
...
+  let bytes = count * sizeof(T)
+  p.toMemAddress().offset(bytes).toPtr(type p[])
+
+template distance*(a, b: MemAddress): int =
+  cast[int](cast[uint](b) - cast[uint](a))
+
+template distance*(a, b: pointer): int =
+  # Number of bytes between a and b - undefined behavior when difference exceeds
+  # what can be represented in an int
+  a.toMemAddress().distance(b.toMemAddress())
```
jangko: no mixin?

stew/ptrops.nim
```diff
@@ -0,0 +1,71 @@
...
+
+template distance*[T](a, b: ptr T): int =
+  # Number of elements between a and b - undefined behavior when difference
+  # exceeds what can be represented in an int
+  {.checks: off.}
+  mixin distance, toMemAddress, toPointer
```
jangko: do we need `toPointer` mixin here?

stew/ptrops.nim
```diff
@@ -0,0 +1,71 @@
...
+
+  a.toMemAddress().distance(b.toMemAddress()) div sizeof(T)
+
+proc `<`*(a, b: MemAddress): bool =
```
jangko: what harm using borrow pragma?

stew/ptrops.nim
arnetheduck: removed

stew/ptrops.nim
arnetheduck: fixed

stew/ptrops.nim
arnetheduck: fixed

stew/ptrops.nim
arnetheduck: could probably in this specific case, though it's on my list of untrusted/buggy features (probably in combination with generics) so I just avoid it

stew/ptrops.nim
```diff
@@ -0,0 +1,70 @@
...
+# been offset is unsafe and many of the operations herein have undefined
+# and platform-specific behavior in corner cases.
+
+# Due to poor codegen and subsequent lack of inlining, many of these operations
+# are templates even where they could be func.
+
+# ByteAddress in std lib is signed - this leads to issues with overflow checking
+# when address is on boundary.
+type
+  MemAddress* = distinct uint
```
zah: What is the semantic difference between a `pointer` and a `MemAddress`? I would say they are the same and this module could be simplified to avoid converting back-and-forth between them.

arnetheduck: you're right, as is it doesn't make much sense. it's a bit of a stepping stone to fix `ByteAddress` which is really broken, but we could remove it - or as an alternative, remove `offset` and friends overloads for `pointer`.

stew/ptrops.nim
```diff
@@ -0,0 +1,54 @@
...
+  ## exceeds what can be represented in an int
+
+  # We assume two's complement wrapping behaviour for `uint`
+  cast[int](cast[uint](b) - cast[uint](a))
+
+template distance*[T](a, b: ptr T): int =
+  # Number of elements between a and b - undefined behavior when difference
+  # exceeds what can be represented in an int
+  {.checks: off.}
+  mixin toMemAddress, distance
```
zah: Ditto

stew/ptrops.nim
```diff
@@ -0,0 +1,54 @@
...
+  mixin offset
+  let bytes = count * sizeof(T)
+  cast[ptr T](cast[pointer](p).offset(bytes))
+
+template distance*(a, b: pointer): int =
+  ## Number of bytes between a and b - undefined behavior when difference
+  ## exceeds what can be represented in an int
+
+  # We assume two's complement wrapping behaviour for `uint`
+  cast[int](cast[uint](b) - cast[uint](a))
```
zah: It would be nice to have a test case for the the wrap-around behavior.

arnetheduck: https://github.com/status-im/nim-stew/pull/8/files#diff-882a0e21f94fbad98903d8677c41e38fR50

stew/ptrops.nim
arnetheduck: removed

PR #10, opened 2019-10-02
Prepare for Nim v1.0.2

zah:


stew/bitops2.nim
```diff
@@ -61,7 +61,13 @@ func firstOneNim(x: uint64): int =
   ## Returns the 1-based index of the least significant set bit of x, or if x is zero, returns zero.
   # https://graphics.stanford.edu/%7Eseander/bithacks.html#ZerosOnRightMultLookup
 
-  if uint32(x) == 0:
+  template convert[T](x: uint64): T =
+    when nimvm:
+      T(x and high(T))
```
arnetheduck: hm, why not always use this code? is the compiler not smart enough to optimize away the bounds check?

jangko: well, I never assume _all_ compiler is smart enough.

arnetheduck: yeah, but also take into account the increased complexity of the code - the separate code paths etc, in this particular case, it's a small difference, but many small differences add up, so generally its better to err towards simplicity unless it's a proven choke point (such as a specific compiler we're interested in emitting bad code)

PR #13, opened 2019-12-10
set, clear, toggle and change bits

arnetheduck:
* use standard bit manipulation naming
* fix signed overflow on large shifts
* add toggle
* add smoke tests
* branch-free changeBit


zah:
These modules are used in the nim-eth/trie library. Please make sure that nothing breaks there.

arnetheduck:
well, all those bitops that were covered by tests work exactly as before ;) 

that said, `nim-beacon-chain` and `nimbus` test suites pass with this change..

PR #14, opened 2019-12-10
add toBytes

arnetheduck:


stew/byteutils.nim
```diff
@@ -123,3 +123,9 @@ func toHex*[N: static[int]](ba: array[N, byte]): string {.inline.} =
   ## Output is in lowercase
   ## No "endianness" reordering is done.
   toHexAux(ba)
+
+func toBytes*(s: string): seq[byte] =
+  ## Convert a string to the corresponding byte sequence - since strings in
+  ## nim essentially are byte sequences without any particular encoding, this
+  ## is almost a noop
```
zah: Actually, this would produce a copy (allocation + copying the range of bytes). Was this your expectation?

arnetheduck: yeah, that's the `almost` - but I keep thinking that it's probably better to avoid this trick altogether and do an explicit init+copy.. hmm.. `string` in nim is such a useless type, might as well have been a possibly distinct alias for `seq[byte]` and life would be easier

mratsim: Ah didn't catch this, we can replace it by a template for no copy

zah: I've advocated for merging strings and sequences myself. The main problem is the null terminator in the strings. Facebook tried to make a clever C++ library where the null terminator is inserted on demand when you call `c_str()`.

PR #18, opened 2020-03-04
non-cast implementation of toBytes, add fromBytes

arnetheduck:


stew/byteutils.nim
```diff
@@ -127,8 +141,12 @@ func toHex*[N: static[int]](ba: array[N, byte]): string {.inline.} =
...
   ## Convert a string to the corresponding byte sequence - since strings in
   ## nim essentially are byte sequences without any particular encoding, this
-  ## is almost a noop
-  cast[seq[byte]](s)
+  ## simply copies the bytes without a null terminator
+  @(s.toOpenArrayByte(0, s.high))
+
+func fromBytes*(T: type string, v: openArray[byte]): string =
+  result = newString(v.len)
+  copyMem(addr result[0], unsafeAddr v[0], v.len)
```
mratsim: I would also add the following (adapted) overload (from https://github.com/status-im/nim-http-utils/pull/9/files)

```Nim
proc toString(data: openarray[byte], start, stop: int): string =
  ## Slice a raw data blob into a string
  ## This is an inclusive slice
  ## The output string is null-terminated for raw C-compat
  let len = stop - start + 1
  result = newString(len)
  copyMem(result[0].addr, data[start].unsafeAddr, len)
```

Slicing an openarray to obtain another openarray is quite verbose at the moment
```Nim
string.fromBytes(v.toOpenArray(start, stop-1))
```
vs an overload
```Nim
string.fromBytes(v, start, stop-1)
```

arnetheduck: `v` for `value` I think which I've seen quite a lot - I just copypasted from somewhere, seems like a good enough name for something generic like this... src would be equally good I guess.

no strong opinion on slicing - this PR is merged already though.. 

your overload however is not safe however: it will overflow undetected if stop is beyond data.len (which is why I don't like `copyMem` in the implementation - programmers can't be trusted with it)

yglukhov: Check for 0 len, otherwise will raise defect for empty array

arnetheduck: ah right, it's not `c` any more ;) thanks @yglukhov

PR #19, opened 2020-03-20
Reviewed stew.result

zah:


stew/result.nim
```diff
@@ -151,12 +151,14 @@ type
...
 func raiseResultError[T, E](self: Result[T, E]) {.noreturn.} =
+  mixin toException
+
   when E is ref Exception:
     if self.e.isNil: # for example Result.default()!
       raise ResultError[void](msg: "Trying to access value with err (nil)")
     raise self.e
-  elif compiles(self.e.toException()):
-    raise self.e.toException()
+  elif compiles(toException(self.e)):
```
zah: This change is probably not strictly necessary, but I've learned to mistrust the use of the dot operator in generic procs.

stew/result.nim
```diff
@@ -232,20 +237,25 @@ template `and`*(self: Result, other: untyped): untyped =
...
   ## Evaluate `other` iff not self.isOk, else return self
   ## fail-fast - will not evaluate other if a is a value
+  ##
+  ## TODO: This API is unsafe due to potential multiple
+  ## evaluation of the `self` parameter.
   if self.isOk: self
   else: other
 
-template catch*(body: typed): Result[type(body), ref Exception] =
+template catch*(body: typed): Result[type(body), ref CatchableError] =
```
zah: Sync `catch` is a synchronous mechanism, I think it shouldn't stop the propagation of non-recoverable errors by default. There might be another API called `catchAllExceptions` that captures defects as well.

stew/result.nim
```diff
@@ -232,20 +237,25 @@ template `and`*(self: Result, other: untyped): untyped =
...
     R.ok(body)
-  except:
-    R.err(getCurrentException())
+  except CatchableError as e:
+    R.err(e)
 
-template capture*(T: type, e: ref Exception): Result[T, ref Exception] =
-  type R = Result[T, ref Exception]
+template capture*(T: type, someExceptionExpr: ref Exception): auto =
+  type ExceptionType = type(someExceptionExpr)
```
zah: This code may be a bit tricky to understand. 

The main idea is to preserve the type of the input exception. Since this is a template, even though the signatures says `ref Exception`, we are still able to obtain the precise type here, so we can return a `Result` with the correct type.

The `let e = someExceptionExpr` line is needed to avoid the double evaluation of the input parameter.

stew/result.nim
```diff
@@ -269,6 +279,32 @@ func `==`*(lhs, rhs: Result): bool {.inline.} =
 func get*[T: not void, E](self: Result[T, E]): T {.inline.} =
   ## Fetch value of result if set, or raise error as an Exception
   ## See also: Option.get
+  ##
+  ## TODO `raiseResultError` is considered a CatchableError which
```
zah: I consider this a serious design issue that needs addressing.

stew/result.nim
arnetheduck: this code is pre-`CatchableError`, so this is indeed a good update - the question perhaps is what to do with the rest of the raw `Exception` support in the library - one way could be to have special "wrapping" or migration code for stuff that inherits from `Exception` but not `Catchable`. same thing with `when E is ref Exception` above - should it limit itself to CatchableError? if yes, the exception list of `raiseResultError` could be bounded better.


I'd generally consider all use of `Defect` UB in terms of behaviour and implementation (ie see manual and the latest `panic` patch) thus it seems wrong to cement "support" for it here by adding `catchAllExceptions` - or if we do add it, perhaps it should be `deprectated` from the start to mark it a support for dealing with legacy code more than anything else.

stew/result.nim
```diff
@@ -254,8 +264,8 @@ template capture*(T: type, e: ref Exception): Result[T, ref Exception] =
     if true:
       # I'm sure there's a nicer way - this just works :)
       raise e
-  except:
-    ret = R.err(getCurrentException())
+  except ExceptionType as catched:
```
arnetheduck: `caught` :)

stew/result.nim
arnetheduck: right, good point - I also forgot to document this one it seems.. 

where would the double evaluation happen? I only see one evaluation, unless you count `typeof`?

zah: Ah, right, I first tried to refactor the code to use `system.getStackTraceEntries` instead of the `raise` trick and this needed the extra variable, but ultimately, it didn't work out and I restored the original version.

stew/result.nim
zah: If we agree to raise a `Defect` here, I think the `toException` feature doesn't make much sense anymore in this particular context. If a `tryGet` function is added, then it may be used to turn the recoverable error value into a recoverable exception.

arnetheduck: yeah, let's do `Defect`.

re `toException`, maybe there is utility still - it gives the user of the library the possibility to customize `Result` for their type as a way to easily transition to the catchable Exception world and it's opt-in. I imagine that if I was converting an API from exceptions to Result, I'd include a backwards-compatible API implemented with `toException` - a little like `boost asio` offers both exception and non-exception-based API. 


One thing that's missing in nim in generalare push and pull conversion feature between error types (enums of different categories) - rust has its general purpose conversion framework with the to/from traits while c++ has constructors and type operators - these help keep the boilerplate down somewhat - `toException` is a special case here but what I'd really want the library to offer is some simple conversions that can be used at abstraction boundaries.

sinkingsugar:
I would like to use `Result` in libp2p, do you recall I should wait this PR or the API anyway is already fine?

arnetheduck:
the changes in here should be relatively harmless unless you rely on `[]` to raise a catchable exception - this is the biggest change - but arguably we don't do that a lot. the branch lgtm mostly though I expect some further small adjustments down the road as we start using it more heavily

stew/result.nim
arnetheduck: so playing around with this a bit more, using `doAssert` for the defect means the defect is not tracked - I have a local version that raises a `ResultDefect` instead (with the error in it, if it's printable).. not sure what will happen with the suggested panic mode though (vs sysfatal)

stew/result.nim
```diff
@@ -175,7 +170,7 @@ type
 func raiseResultError[T, E](self: Result[T, E]) {.noreturn.} =
   mixin toException
 
-  when E is ref Exception:
+  when E is ref Defect or E is ref CatchableError:
```
arnetheduck: this change looks a bit overzealous - I mean, if someone's wrapping legacy code, they'll have a bunch of exceptions around, and we're not really inheriting from it..

sinkingsugar:
I have a question I asked on discord but had no answers:

> how would you check exception type with readError? `isnot` and `of` operators are failing me
and I want to avoid cstring comparison of name
also reraise trick
but I guess I have to re-raise .. cos is and `typeof(exc)` of are not working
this applies to `Result` too
I don't see any such test case either in result/stew

arnetheduck:
> how would you check exception type with readError? isnot and of operators are failing me

what's `readError`? exceptions are dynamically typed objects and typically you add `except` clauses for every type you want to catch

> cstring comparison of name

strings in errors in general is something you use when you want to provide diagnostic information without the expectation that calling code will have logic to differentiate between different types of errors - this is what happens with most error handling in practise if your API is well written (ie performs one logical "task" that either works or fails

> but I guess I have to re-raise

the thing you normally want to do with `Result` is to handle errors from the layer below within your layer and translate it to something appropriate for your abstraction - ie not let errors from deep with leak through the abstraction - so often you'll do some sort of translation or simplification and not straight-off forwarding of errors

sinkingsugar:
Hmmm, I meant basically this
(Copy pasting from some rust)

```rust
use std::io;
use std::num;

// We derive `Debug` because all types should probably derive `Debug`.
// This gives us a reasonable human-readable description of `CliError` values.
#[derive(Debug)]
enum CliError {
    Io(io::Error),
    Parse(num::ParseIntError),
}
```

and handle like
```rust
use std::error;
use std::fmt;

impl fmt::Display for CliError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match *self {
            // Both underlying errors already impl `Display`, so we defer to
            // their implementations.
            CliError::Io(ref err) => write!(f, "IO error: {}", err),
            CliError::Parse(ref err) => write!(f, "Parse error: {}", err),
        }
    }
}

impl error::Error for CliError {
    fn cause(&self) -> Option<&error::Error> {
        match *self {
            // N.B. Both of these implicitly cast `err` from their concrete
            // types (either `&io::Error` or `&num::ParseIntError`)
            // to a trait object `&Error`. This works because both error types
            // implement `Error`.
            CliError::Io(ref err) => Some(err),
            CliError::Parse(ref err) => Some(err),
        }
    }
}
```

I know we don't have match, but at least I would like to avoid `re-raise` and instead check the type with `of` or `is`. But I guess it's not possible.
Basically treat it as a value.
```nim
if err of MyBeautifulFailure:
  dosomething()
elif err of MyUglyFailure:
  exit(99)
```

arnetheduck:
```
try:
  raise (ref ValueError)(msg: "test")
except CatchableError as e:
  doAssert e of (ref ValueError)
```

exceptions are ref, need to include that in your type check

sinkingsugar:
I know they are refs and all but yeah exactly the snipped you made is what I wanted to avoid for readability.

Also to separate from exceptions mind set. The code you wrote is clear for us but anyone reading will think, are we using result or exceptions? Do I have to catch stuff?

sinkingsugar:
Ah sorry I saw the of after.
That did not compile for me. On a `ref Exception`.

arnetheduck:
what didn't compile?

I mean:
```
var err: Exception = newException(ValueError, "test")
if err of (ref ValueError):
  echo "hello"
```
compiles, as does the example I posted

arnetheduck:
if you mean unwrapping and pattern-matching, then I don't know of any existing construct, but there's a library that @alehander92 was working on that help do some pattern matching at least - not sure of it takes dynamic type into consideration.. 

anyway, I'd suggest starting with imperfect constructs to build up a corpus of how result can be used, then we'll see what helpers need to be added to make it smooth to use

PR #20, opened 2020-03-28
restore bridge mode for exceptions

arnetheduck:


arnetheduck:
re #19

arnetheduck:
the intent here is to experiment further with bridge mode, and maybe remove it as #19 implicitly suggests - in the meantime, we should "normally" not be raising catchable errors from Result

PR #21, opened 2020-03-29
stew/objects: add baseType()

stefantalpalaru:


stew/objects.nim
```diff
@@ -31,3 +31,15 @@ when not compiles(len((1, 2))):
...
     arity(type(x))
 
+# Get an object's base type, as a cstring. Ref objects will have an ":ObjectType"
+# suffix.
+# From: https://gist.github.com/stefantalpalaru/82dc71bb547d6f9178b916e3ed5b527d
+proc baseType*(obj: RootObj): cstring =
+  when not defined(nimTypeNames):
+    raiseAssert("you need to compile this with '-d:nimTypeNames'")
+  else:
+    {.emit: "result = `obj`->m_type->name;".}
```
arnetheduck, line 41: the C code that nim emits is unstable between versions - this is very hacky, to say the least. what would alternatives be?

stefantalpalaru: No alternatives, while staying outside the stdlib. That C struct has been stable for a long time and I don't think there are any plans to change it.

arnetheduck: well, can you petition upstream for a proper fix? this isn't an unusual problem..

stefantalpalaru: It's not a fix, it's a new feature and that means it can stew for a year or so in a Nim development branch, if I manage to get it merged in the first place. Let it stew here, instead :-)

arnetheduck: well, this breaks wasm, js, nlvm etc already - it probably breaks with the new arc runtime as well, which is the point where RTTI changes

PR #22, opened 2020-03-30
result: cleanups

arnetheduck:
* fix ResultError type
* add/fix documentation
* clean up raises (in preparation for better Defect handling)
* fix toException mixin
* work around compiler bug with more explicit types
* fix capture exception type
* fix result type on `?`

these are the non-breaking changes in #19 / #20

### DIFF
--- a/stew/result.nim
+++ b/stew/result.nim
@@ -7,10 +7,10 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 type
-  ResultError*[E] = ref object of ValueError
+  ResultError*[E] = object of ValueError
     ## Error raised when trying to access value of result when error is set
     ## Note: If error is of exception type, it will be raised instead!
-    error: E
+    error*: E
 
   Result*[T, E] = object
     ## Result type that can hold either a value or an error, but not both
@@ -19,7 +19,7 @@ type
     ##
     ## ```
     ## # It's convenient to create an alias - most likely, you'll do just fine
-    ## # with strings as error!
+    ## # with strings or cstrings as error
     ##
     ## type R = Result[int, string]
     ##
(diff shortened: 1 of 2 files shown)


stew/result.nim
```diff
@@ -151,40 +173,42 @@ type
...
   R(o: true, v: x)
 
-template ok*(self: var Result, x: auto) =
+template ok*[T, E](self: var Result[T, E], x: auto) =
   ## Set the result to success and update value
   ## Example: `result.ok(42)`
   self = ok(type self, x)
 
-template err*(R: type Result, x: auto): auto =
+template err*[T, E](R: type Result[T, E], x: auto): R =
```
zah, line 200: Believe it or not, but the short version of the code will be actually more robust (the generic param binding in templates is currently broken) and it will compile slightly faster on every call-site :)

stew/result.nim
```diff
@@ -220,45 +244,62 @@ func mapCast*[T0, E0](
...
-  except:
-    ret = R.err(getCurrentException())
+      raise someExceptionExpr
+  except E as caught:
+    ret = R.err(caught)
   ret
 
-func `==`*(lhs, rhs: Result): bool {.inline.} =
+func `==`*[T0, E0, T1, E1](lhs: Result[T0, E0], rhs: Result[T1, E1]): bool {.inline.} =
   if lhs.isOk != rhs.isOk:
```
zah, line 303: Interesting level of flexibility here, I think you need to mixin `==` to make it work correctly.

stew/result.nim
arnetheduck: Believe it or not, but the nim compiler has bugs: https://github.com/nim-lang/Nim/issues/13799

arnetheduck: (yes, devel too)

stew/result.nim
arnetheduck: aren't mixins for templates?

the flexibility is for the same reason you want heterogeneous lookup in maps etc - ie `Result` should not have an opinion here

zah: `mixin` is for both generics and templates, but it has some tricky rules. If it sees an overloaded function in the current scope, it considers the symbol name as mixed in by default. This rule should probably kick in for the `==` operator actually, but it may be a good habit to always state your intentions explicitly.

PR #23, opened 2020-03-31
result: add `expect` helper

arnetheduck:


mratsim:
One caveat though, it will probably conflict with unittest https://nim-lang.org/docs/unittest.html#expect.m%2Cvarargs%5Btyped%5D%2Cuntyped

Though I think the issue is on the macro signature not being restricted enough

i.e.
```Nim
macro expect(exceptions: varargs[typed]; body: untyped): untyped
```

instead of

```Nim
macro expect(exceptions: varargs[Exception]; body: untyped): untyped
```

PR #25, opened 2020-04-05
result: fix and/or template evaluation, add perf notes

arnetheduck:


stew/result.nim
```diff
@@ -171,16 +171,51 @@ type
     ## * Pattern matching in rust allows convenient extraction of value or error
     ##   in one go.
     ##
+    ## # Performance considerations
+    ##
+    ## Compared to a simple return type, returning a Result sometimes incurs
+    ## an overhead. Compared to raising an exception, this overhead is very low.
```
zah, line 177: My understanding is that the comparison with exceptions is more subtle.

In the success path, exceptions will be the cheapest error-handling mechanism if we compile in C++ mode with a 64-bit target. The evidence for this is coming from a well-known C++ performance report:
http://www.open-std.org/jtc1/sc22/wg21/docs/TR18015.pdf

In case of failure though, exceptions will be much more costly than result (orders of magnitude).

I've never seen thorough analysis on the C compilation mode with `setjmp` (only some Java papers that were comparing the different ways to implement exceptions in the Jikes RVM, which may not be quite representative).

The new `--exceptions:goto` option may bring `Result` and exceptions much closer in both scenarios, but it will preserve optimizations such as RVO.

arnetheduck: I'm not sure what you mean - this is what the text says: generally more expensive than returning a simple value (in the ok case implied), generally less expensive than raising an exception (in the error case implied).

The former is qualified with "sometimes" because in simple value cases (ie `Result[int]`) and handling the error directly (for example compared to `try/catch` of parsing `int`), there is no practical difference and the `Result` code might just as well be faster since it's more compact and can be inlined more easily - this is what exceptions:goto exploits to a certain degree, and also why swift does it this way - the difference between passing an exception around using dwarf eh and explicitly passing it up the stack in a register is negligeable to a certain extent, with the latter more optimizeable from a control-flow point of view and the former better for making big jumps.

Can you point to a specific page in that report? It's a bit long, and from 2006 - quite a few things have happened on the compiler and cpu fronts since then.

Also, taking a step back, I don't think the performance of these things matter that much tbh - the tend to be in the range of a few % with a pretty large variance depending on the kind of code meaning that how you write your code will have a bigger impact than the specific mechanism you use. Take `rust` for example - it's generally accused of being overly strict on the programmer, but rarely _slow_. Of course you can find egregiously slow cases of using `Result` - specially if you go nuts and use it everywhere down to the last addition - there, `exceptions:goto` (but without that beefy exception object - instead just a light value type like an enum kind of flag) would really shine.

I look at this whole debate a lot more from an API clarity point of view - the performance difference in cases where it matters can usually be sorted with "alternative" techniques like restructuring the code to fit the error handling model better - you can write slow code with both EH and exceptions if you want.

arnetheduck: > I've never seen thorough analysis on the C compilation mode with setjmp

this misses the point though, a little - the missing comparison here is "return a hidden flag explicitly in every function" which is `exceptions:goto` or, pragmatically and more explicitly, `Result`.

for example, with `exceptions:goto` you can get one "nesting" level of RVO back, but you have to compare that with the memory allocation of the exception object and other exception code overhead - and you should be comparing with a "move" implementation of `Result` in which copies are assumed to be cheap - ie RVO makes sense for when copies are expensive, but if they're not, it doesn't really matter.

in fact, most `Result` problems stem from this fact: it copies too much and the language doesn't allow us to avoid the copies (yet).

zah: The discussion of error handling starts at page 32 in the report. 

In essence, the 32-bit compilers of the past tended to use book-keeping execution code to keep track of where we are, what objects need to be destroyed during stack unwinding and so on. The modern compilers replace this mechanism with a "table-based" approach where all unwinding logic is derived from the program counter location when the exception is being thrown. There is no book-keeping code and more importantly **no error-checking code** during normal execution. The tables that are necessary for this tend to be large and complicated, which makes triggering the unwinding mechanism more expensive in return of faster execution in the happy path.

I don't think much have changed in this space, I often see the report cited in recent discussions.

The significant cost of raising exceptions may be considered a potential DoS mechanism though. One has to be careful about it, especially if there is more sophisticated error reporting and exceptions-statistics-gathering in place.

arnetheduck: Alright, rewrote this to point out some of the specific overheads as a guide for writing efficient Result-based code - in general, I'm starting to feel that an extended discussion of relative merits at a language level might be beyond what is good use of space here.

PR #28, opened 2020-04-09
result: options compatibility

arnetheduck:
motivating example for `contains`:

```
if IndexError in res.err():  # calls error-to-option then `contains`
  ...
```

### DIFF
--- a/stew/results.nim
+++ b/stew/results.nim
@@ -573,3 +573,32 @@ template `?`*[T, E](self: Result[T, E]): T =
   if v.isErr: return err(typeof(result), v.error)
 
   v.value
+
+# Options compatibility
+
+import options
+export options
+
+when defined(resultsExperimental):
+  {.pragma: resexp.}
+else:
+  {.pragma: resexp, deprecated: "experimental".}
+
+func ok*[T, E](r: Result[T, E]): Option[T] {.resexp.} =
+  if r.isOk():
+    return some(r.v)
+
+func err*[T, E](r: Result[T, E]): Option[E] {.resexp.} =
+  if r.isErr():
+    return some(r.e)
+
+func mapErr*[T, E](o: Option[T], f: proc (): E): Result[T, E] {.resexp.} =
+  if o.isSome():
+    ok(o.get())
+  else:
+    err(f())
+
+# Options extras
+
+func contains*[T](o: Option[T], v: T): bool {.resexp.} =
+  o.isSome() and o.get() == v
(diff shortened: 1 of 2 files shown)


stew/results.nim
```diff
@@ -573,3 +573,32 @@ template `?`*[T, E](self: Result[T, E]): T =
...
+
+import options
+export options
+
+when defined(resultsExperimental):
+  {.pragma: resexp.}
+else:
+  {.pragma: resexp, deprecated: "experimental".}
+
+func ok*[T, E](r: Result[T, E]): Option[T] {.resexp.} =
```
zah, line 587: I find the naming of this somewhat confusing. I use `(var result).ok` to set the result to the OK state, but calling `ok()` gives me an `Option`?

arnetheduck: yeah I thought about that, though strictly it's `res.ok()` and `res.ok(value)` - it reads kind of nice when doing the `in` trick - I'm more sceptical about `(var res).ok(value)` to be honest - it's there because we don't have the nice enum constructors where you only need to know either T or E - suggestions?

zah: It's just `result.ok()` in the `void` case, but then `Option[void]` is probably pointless. 
Nevertheless, why don't you go with a more obvious name such as `getOpt`?

zah: I tend to use the following syntax in my own code:

```nim
return ok foo()
...
return err "bar"
```

Or the versions without `return` where that's appropriate, so indeed the `var` overloads may be less important.

arnetheduck: `getOpt` / `toOption` no because both `ok` and `err` return an option so it's a bit ambiguous - also, it fits `isOk`/`isErr` - ie everything is within the same "name" world.

the `var` overloads are there for when you're not returning, but using `Result` outside of a return value, really, and `typeof(result)` is meaningless or wrong.

also, `if X in res.toOption` doesn't look as nice.

arnetheduck: the void case, that's a good point. hm.

stefantalpalaru: `toOption` makes sense to me, because I don't have to guess which `ok` variant is executed there.

arnetheduck: `setOk`/`setErr` for `var`? or remove them indeed and readd when there's a good use case

zah: But when is the `toOption` construct useful? Why are we giving it more priority by assigning the shortest possible name?

arnetheduck: well, options and results are used in similar situations, so they complete each other in a way - it's often useful to move between the two in a friction-less way.

but the main reason for not liking the `toOption` name is that it biases the interface towards the `ok` case which a bit defeats they purpose of `Result`.

arnetheduck:
superceded by #177

PR #29, opened 2020-04-09
hex byte: raise instead of assert on hex-to-byte odd lengths

arnetheduck:


stew/byteutils.nim
```diff
@@ -62,8 +62,8 @@ func hexToByteArray*(hexStr: string, output: var openArray[byte], fromIdx, toIdx
   doAssert(fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len)
   let sz = toIdx - fromIdx + 1
 
-  doAssert hexStr.len - sIdx >= 2*sz
-
+  if hexStr.len - sIdx < 2*sz:
```
zah, line 65: Nimcrypto saw a similar fix, but there we require the `hexStr` to precisely match output size. I'm not sure which policy is better.

arnetheduck: hm, now that I think about it, the stricter version probably makes more sense in this case - the caller won't know how many bytes were actually written which I imagine might be an issue..

arnetheduck: ok, and then I thought about it again, and the caller passes in the indices, so I guess they know what they're doing... the whole api is odd though, would be nice to have a slice operator for openarray that made sense. I'd leave it as-is for now.

arnetheduck: I have a slightly longer term idea to split out the hex stuff into a separate file (`base16` as @cheatfate suggested) - can rethink the API then - until then, there's even a unit test that tests that the behaviour is this so maybe there's a reason (and thus, changing would probably deserve a separate PR)

PR #30, opened 2020-04-15
change get/[] to always raise Defect and tryGet to do eh bridge mode

arnetheduck:
* also sprinkle mixin randomly across the codebase


arnetheduck:
this should complete #19 - see discussion there

PR #33, opened 2020-04-22
results.Opt and objects.declval

zah:
Please see the commit comments.

### DIFF
--- a/stew/objects.nim
+++ b/stew/objects.nim
@@ -42,6 +42,8 @@ func declval*(T: type): T {.compileTime.} =
   ## 2. It will work for types that lack a valid default value due
   ##    to `not nil` or `requiresInit` requirements.
   ##
+  doAssert false,
+    "declvar should be used only in `typeof` expressions and concepts"
   default(ptr T)[]
 
 when not compiles(len((1, 2))):
--- a/stew/objects.nim
+++ b/stew/objects.nim
@@ -25,6 +25,25 @@ template anonConst*(val: untyped): untyped =
   const c = val
   c
 
+func declval*(T: type): T {.compileTime.} =
+  ## `declval` denotes an anonymous expression of a particular
+  ## type. It can be used in situations where you want to determine
+  ## the type of an overloaded call in `typeof` expressions.
+  ##
+  ## Example:
+  ## ```
+  ## type T = typeof foo(declval(string), declval(var int))
+  ## ```
+  ##
+  ## Please note that `declval` has two advantages over `default`:
+  ## 
+  ## 1. It can return expressions with proper `var` or `lent` types.
+  ## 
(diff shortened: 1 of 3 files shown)


stew/objects.nim
```diff
@@ -25,6 +25,25 @@ template anonConst*(val: untyped): untyped =
...
+  ## ```
+  ##
+  ## Please note that `declval` has two advantages over `default`:
+  ## 
+  ## 1. It can return expressions with proper `var` or `lent` types.
+  ## 
+  ## 2. It will work for types that lack a valid default value due
+  ##    to `not nil` or `requiresInit` requirements.
+  ##
+  default(ptr T)[]
```
arnetheduck, line 45: why `default` in here - and why make it do something reasonable? if  it's used in typeof, the compiler doesn't actually run the code inside (or?) so it could just raise an exception or something (in case someone uses it in an evaluated context) - at least, this is the case for c++ typeof..

zah: If `T` is a type without a default value (e.g. because it has a `not nil` field or it's `requiresInit` type), then the compiler will produce a warning if the body was empty. You need to return something and this pointer trick is how you fool the compiler.

arnetheduck: ok - well, I'd guess the risk here is that I do something like `static: declval(MyType)` - I was hoping that a raise would give a good error message in that case

arnetheduck: ah, and I understood the `default` wrong on first read as well - I read `default(T)` and not `default(ptr T)` - I guess it's just a fancier way of saying `nil` - any.. advantage?

zah: `default(T)` is also illegal for `requiresInit` types. Dereferencing a nil pointer is not (yet).
The latest version of the code will now assert if you accidentally try to execute the function.

arnetheduck: my question was: why not use `nil`? `default(ptr T) == nil` so `default` is just an obfuscation

zah: `nil` doesn't have a type. The full version of the code would have to be `cast[ptr T](nil)[]` which is too scary for such a safe feature :)

arnetheduck: why? `(ptr T)(nil)` works?

arnetheduck:
ugh, can you add some `[T, void]` tests? `Result` has had pretty good coverage so far..

stew/objects.nim
zah: Beauty is in the eye of the beholder. For me, `default(ptr T)[]` is more clear than `(ptr T)(nil)[]`. Wins by a number of braces :)

arnetheduck: loses by keystrokes and logical indirections ;) I got my assert at least.

PR #43, opened 2020-06-15
fix bitops2 checkedScan and bitScan for vcc

jangko:


arnetheduck:
this is not covered by tests?

jangko:
our appveyor configuration using mingw-gcc, not vcc. so the answer is yes, all vcc related code are not tested.

zah:
It should be possible to add Visual Studio targets though:
https://www.appveyor.com/docs/lang/cpp/#visual-studio

jangko:
one thing I hate about Nim default `nim.cfg` regarding vcc is Nim will call vcc by a proxy executable called `vccexe.exe` which slow down compilation tremendously.

PR #44, opened 2020-06-22
Add 'checkedEnumAssign' for validating tainted enum inputs

zah:


stew/objects.nim
```diff
@@ -64,3 +67,43 @@ proc baseType*(obj: RootObj): cstring =
...
+  #    Sym "A2"
+  #    Sym "B2"
+  #    Sym "C2"
+  #
+  macro hasHoles*(T: type[enum]): bool =
+    let t = getType(T)[1]
+    echo t.treeRepr
+    return newLit(true)
+
+func checkedEnumAssign*[E: enum, I: SomeInteger](res: var E, value: I): bool =
```
arnetheduck, line 91: meh, what's the point of adding this when it has a pretty significant gotcha? too easy to use the wrong way

zah: We'll need this for any serialization format that supports enums.

arnetheduck: yes, but the caller won't even know their code doesn't work until it raises a defect in some edge case, which defeats the purpose of the helper - better have them write the code manually and think about their enums - or make this function warn somehow about gap support at compile time

PR #48, opened 2020-07-06
Fix #47

### LINKED ISSUE #47: Templates in ptrops.nim inserting {.checks: off.} into all files that use them
Hey, just reporting something I noticed while investigating why some code was changing its behavior when this file was imported. I found that these templates have {.checks: off.} which results in that being inserted anywhere these functions are used in code.

It doesn't seem great to potentially turn checks off at the global / program scope just by using a little helper from a library. Could this push and pop the checks:off or is there a good reason for this I'm not aware of?

```
template offset*[T](p: ptr T, count: int): ptr T =
  ## Offset a pointer to T by count elements. Behavior is undefined on
  ## overflow.

  # Actual behavior is wrapping, but this may be revised in the future to enable
  # better optimizations.

  # We turn off checking here - too large counts is UB
  {.checks: off.}
  let bytes = count * sizeof(T)
  cast[ptr T](offset(cast[pointer](p), bytes))

template distance*[T](a, b: ptr T): int =
  # Number of elements between a and b - undefined behavior when difference
  # exceeds what can be represented in an int
  {.checks: off.}
  distance(cast[pointer](a), cast[pointer](b)) div sizeof(T)```

zah:
As far as I can tell, this preserves the intended lack of checks in these low-level routines while dealing with the problem explained in #47. By default, Nim insert checks only for `int` operands. The `uint` arithmetic used in `offset` and `distance` doesn't introduce any checks.

### DIFF
--- a/stew/ptrops.nim
+++ b/stew/ptrops.nim
@@ -36,8 +36,9 @@ template offset*[T](p: ptr T, count: int): ptr T =
   # better optimizations.
 
   # We turn off checking here - too large counts is UB
-  {.checks: off.}
+  {.push checks: off.}
   let bytes = count * sizeof(T)
+  {.pop.}
   cast[ptr T](offset(cast[pointer](p), bytes))
 
 template distance*(a, b: pointer): int =
@@ -50,5 +51,5 @@ template distance*(a, b: pointer): int =
 template distance*[T](a, b: ptr T): int =
   # Number of elements between a and b - undefined behavior when difference
   # exceeds what can be represented in an int
-  {.checks: off.}
   distance(cast[pointer](a), cast[pointer](b)) div sizeof(T)
+


arnetheduck:
`count` and `sizeof` are `int` though..

zah:
Yes, but the only dangerous place where these two are used is still wrapped in `{.push checks: off.}`.
Do you object to the removal of the checks in `distance`? `sizeof` is at least 1 for all object types (even empty ones)

arnetheduck:
hm.. there was a reason.. ah, now I remember - pointers are often on the `0x7ffff...` boundary and can tip over, so when done with `int` semantics the checks actually trigger - that goes for the `div` as well afair.

arnetheduck:
hm, tbh I'm unsure if really applies to `div` - or at least I can't figure out how right now, or why I put the check there.

zah:
Already fixed in another PR

PR #49, opened 2020-07-10
AsyncResults (andThen/orElse)

sinkingsugar:
Started as an experiment to see if I could get some similarities to rust tokio `andThen` and `orElse` helpers.
It actually worked without resorting to any sorcery or editing chronos.

I would like to use it in libp2p for example. as it allows exceptions handling free code (the abstractions collects errors).
Obviously it's not meant to replace all the existing code.. just those times were we want a well defined chain of actions with one entry/exit point.

Anyway, opening this PR for feedback/discussion

### DIFF
--- a/stew/asyncresults.nim
+++ b/stew/asyncresults.nim
@@ -1,0 +1,236 @@
+import results
+import chronos # requires chornos
+
+type
+  AsyncResult*[T] = Future[Result[T, string]]
+  AsyncCall*[T1, T2] = proc(prev: AsyncResult[T1]): AsyncResult[T2]
+
+# the whole point is not loose errors around and to be able to declare a chain of async ops clearly
+# we do strict error checking and propagate them in the result
+# and so we need just a single await in user code
+
+proc andThen*[T1, T2](a: AsyncResult[T1]; b: AsyncCall[T1, T2]): AsyncResult[T2] = 
+  var retFuture = newFuture[Result[T2, string]]()
+  a.callback = proc(_: pointer) =
+    if a.failed():
+      let msg = a.readError().msg
+      retFuture.complete(err(Result[T2, string], msg))
+    else:
+      let ares = a.read()
+      if ares.isOk():
+        try:
+          let f = b(a)
+          f.callback = proc(_: pointer) =
+            if f.failed():
+              let msg = f.readError().msg
+              retFuture.complete(err(Result[T2, string], msg))
+            else:
+              let res = f.read()
+              retFuture.complete(res)
+          f.cancelCallback = proc(_: pointer) =
+            f.callback = proc(_: pointer) = discard # empty up callback
+            # since we complete here
+            retFuture.complete(err(Result[T2, string], "Operation canceled"))
+        except CatchableError as ex:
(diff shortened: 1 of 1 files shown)


zah:
Can you show some motivating examples for this? This style of programming tends to be popular in languages before they gain async/await support.

sinkingsugar:
Well, rust async/await codebases tend to have a lot of this anyway because of the nature of the language and error handling.
It still uses async/await btw, this is not callback based... notice that each closure is `async` transformed.

The rationale is to build a chain of async actions explicitly, avoiding the pitfalls of regular async transformed procedures full of awaits (race conditions, forgetting exception handling, etc) and to have clear entry/exit points where data/errors are collected.

It is not meant to replace the common actual style with this btw, this is strictly for certain cases were you wanna be quite sure of the flow and error management. (e.g. libp2p)

zah:
I'm still confused. If I have a chain like `await op1().andThen(op2).andThen(op3)`, how is this semantically different than using multiple awaits?

```nim
let res1 = ? await op1()
let res2 = ? await op2(res1)
let res3 = ? await op3(res2)
```

It seems to me that the race conditions in the code will be exactly the same. The same can be said for the guarantees regarding order of execution. If I had forgotten to put exception handling around the whole block, wouldn't this be a problem for both approaches?

One downside that I see with the `andThen` API is that it relies much more heavily on GC-based closures, while certain implementation strategies for await can avoid the reliance on a GC.

sinkingsugar:
It's indeed sugar and just another way to write and see things that fits certain scenarios.

There is no difference except that you have to write exception handlers and handle them up to pull out an error message to fail plus result possible errors.

But there are a few issues :

* `?` does not work in an async transformed proc.
* async transformed procs return type is checked in the macro by string so cannot be an alias 
* etc 

Race conditions are fixed cos your way of writing code changes, with this approach you think in "blocks"

About closures and GC, First of all that's the language fault, second they are all over our code base anyway ! ( I see tons in NBC, everywhere in libp2p and chronos itself is using them a lot internally ) so I'm sorry but Altho I understand the point I cannot really consider it something to care about at this point.

cheatfate:
About race conditions, how it can fix it. For example you are running two concurrent tasks with your constructions inside, and both your tasks accessing reading/writing one global variable? Race condition is still here, you just trying to introduce explicit so named "callback hell" into "async/await". I strongly oppose this approach.

sinkingsugar:
> I strongly oppose this approach

Fine for me to waste another couple of months finding exceptions

kdeme:
The main benefits here regarding exception handling come from the fact that you use `AsyncResult`.
You would have the same effect with the current style, when using `AsyncResult` (Yes, async `?` doesnt work, it would be nice if that could get fixed). 
I guess that if you use the sugar, you are enforced to use `AsyncResult`, which is some sort of benefit :). 

> Race conditions are fixed cos your way of writing code changes, with this approach you think in "blocks"

Could you explain the benefit regarding race conditions? Do you mean as in not forgetting an await and running something with `asyncCheck` for example?  As it doesn't help for the cases that @cheatfate mentions.

PR #52, opened 2020-08-07
Add tables.mgetOrPutLazy

zah:


stew/shims/tables.nim
```diff
@@ -9,4 +9,15 @@ template init*[A, B](T: type OrderedTableRef[A, B]): auto = newOrderedTable[A, B
...
 
+template mgetOrPutLazy*[A, B](t: Table[A, B], key: A, val: B): var B =
+  type R = B
+
+  proc setter(loc: var R): var R =
+    if loc == default(R):
+      loc = val
+    loc
+
+  setter(mgetOrPut(t, key, default(R)))
```
arnetheduck, line 20: this looks pretty odd - ie if `B` is an `array[10000, uint64]`, this will.. zero-init such an array, zeroinit another such array (`default(R)`) then compare all those values?

stew/shims/tables.nim
```diff
@@ -9,4 +9,15 @@ template init*[A, B](T: type OrderedTableRef[A, B]): auto = newOrderedTable[A, B
 template init*[A](T: type CountTable[A]): auto = initCountTable[A]()
 template init*[A](T: type CountTableRef[A]): auto = newCountTable[A]()
 
+template mgetOrPutLazy*[A, B](t: Table[A, B], key: A, val: B): var B =
+  type R = B
+
+  proc setter(loc: var R): var R =
+    if loc == default(R):
```
arnetheduck, line 16: so if the collection had a value that was all zeroes (as opposed to not having a value), this will update it to `B` which doesn't seem to be the intent

zah: Yes, this version of code comes with a number of caveats, but the idea of having a shim is that it allows us to capture the intention of the user until a more general and efficient version of the API is added to the `tables` module as a template similar to `withValue`.

At the moment it's reasonable to use mostly with tables with `ref` values that don't need to store `nil`. The `default` function for the type should be cheap and not a valid value in the usage context.

arnetheduck: well, the that's the point a bit - my intent is not to set if the value is the default - it's to set / create only if it's not in the collection already

PR #53, opened 2020-08-17
IO2

cheatfate:
io2 module is Result[T] version of `lib/system/io.nim` and `lib/pure/os.nim` where main features are:
1. Do not uses exceptions for error handling.
2. Do not uses `FILE` for file processing.
3. Ability to set permissions for created files using `writeFile()`.
4. Ability to set permissions for creating folders using `createPath()`.

Right now `io2` could not replace full functionality of `io.nim` and `os.nim`, it only replaces limited amount of operations which is required to fix https://github.com/status-im/nim-beacon-chain/issues/1320


stew/safeio.nim
```diff
@@ -0,0 +1,563 @@
+import results
```
arnetheduck, line 1: remove file?

cheatfate: oops.

stew/io2.nim
```diff
@@ -0,0 +1,1003 @@
...
+       importc, header: "<unistd.h>", sideEffect.}
+  proc `==`*(a: IoErrorCode, b: cint): bool {.inline.} =
+    (cint(a) == b)
+
+type
+  IoResult*[T] = Result[T, IoErrorCode]
+
+  OpenFlags* = enum
+    ReadOnly, WriteOnly, ReadWrite, Create, Exclusive, Append, Truncate,
+    NoInherit, NonBlock, Direct
```
alaviss, line 168: I think it'd be better if we use `Read` and `Write` flags instead of having three different flags.

Also for safety reasons I'd prefer having `NoInherit` being the default. The stdlib has already make this change.

alaviss (commented):
It appears to me that a lot of `os` and `winlean` has been duplicated into this module. Is there any reason for this?

stew/io2.nim
cheatfate: @alaviss from my point of view `stdlib` is incorrect because of this, it should not make any decisions on what should be default and what should not. It should allow developers to choose their own way.

cheatfate:
Most of the `winlean` declarations are incorrect or non stable.

1. `int32` as `DWORD`, but `DWORD` is actually `uint32`.
2. Sometimes return values are missed completely.
3. Some of the declarations are made for specific case `useWinUnicode`, so API calls you are using can become unavailable when compilation flags got changed.

cheatfate:
And we are not going to extend `os` or `io` modules of Nim, we going to replace it completely.

stew/io2.nim
alaviss: It's the wrong default. The reason why most operating system API doesn't use `NoInherit` by default is purely historical. They can't change that now due to compatibility reasons, but as a (relatively new) `stdlib` we can, and we did.

Nowadays people have to add syscalls like `closefrom`, `close_range` just to walkaround the historical baggage of not having `NoInherit` (`O_CLOEXEC`) as the default. Even Microsoft has been making all "new" APIs in Windows have inheritance disabled by default (`bInheritHandles` defaults to 0 when not provided for many APIs). And I doubt any of them were opinionated, it's just the reality that FD inheritance was the wrong default, and one that bites.

alaviss:
> Most of the `winlean` declarations are incorrect or non stable.

Opened nim-lang/Nim#15231 to track this. Please consider opening issues regarding problems that you (Status) found within the stdlib, as you are one of the biggest user of Nim, so implementation issues affecting you would likely affect other Nim users.

stew/io2.nim
```diff
@@ -0,0 +1,1007 @@
...
+  ## program; where basename for "/foo/bar/" returns "bar", the basename()
+  ## procedure returns an empty string ("").
+  splitPath(path)[1]
+
+proc dirname*(path: string): string =
+  ## Return the directory name of pathname ``path``.
+  splitPath(path)[0]
+
+when defined(windows):
+  proc toLongPath*(path: string): IoResult[string] =
```
arnetheduck, line 322: hm, feels like a separate type than `string` would be more appropriate here

stew/io2.nim
```diff
@@ -0,0 +1,1007 @@
...
+
+proc readFile*(handle: IoHandle,
+               data: var openarray[char]): IoResult[uint] {.inline.} =
+  ## Reads ``len(data)`` characters from file descriptor ``handle`` and store
+  ## this characters to ``data``.
+  ##
+  ## Returns number of bytes characters read from file descriptor.
+  readFile(handle, data.toOpenArrayByte(0, len(data) - 1))
+
+proc writeFile*(pathName: string, data: openarray[byte],
```
arnetheduck, line 946: it's actually spelled `openArray` in `system`

stew/io2.nim
```diff
@@ -0,0 +1,1007 @@
...
+                createMode: int = 0o644): IoResult[void] =
+  ## Opens a file named ``pathName`` for writing. Then writes the
+  ## content ``data`` completely to the file and closes the file afterwards.
+  ##
+  ## If file is not exists it will be created with permissions mask
+  ## ``createMode`` (default value is 0o644).
+  ##
+  ## If file is already exists, but file permissions are not equal to
+  ## ``createMode`` procedure will change permissions first and only after
```
arnetheduck, line 955: this might be annoying: if user explicitly changes permissions of an existing file and the program resets them, that's not good - I think in most cases keeping the existing permissions is a more reasonable thing to do.

stew/io2.nim
```diff
@@ -0,0 +1,1007 @@
...
+proc writeFile*(pathName: string, data: openarray[char],
+                createMode: int = 0o644): IoResult[void] {.inline.} =
+  ## Opens a file named ``pathName`` for writing. Then writes the
+  ## content ``data`` completely to the file and closes the file afterwards.
+  ##
+  ## If file is not exists it will be created with permissions mask
+  ## ``createMode`` (default value is 0o644).
+  writeFile(pathName, data.toOpenArrayByte(0, len(data) - 1), createMode)
+
+proc readAllFile*(pathName: string, blockSize = 16384'u): IoResult[seq[byte]] =
```
arnetheduck, line 985: perhaps `readAllFile(name: string, buf: var seq[byte): IOResult[void]` to stay in line with the other `readFile`? this overload could then stay around as well.

also, I'm not sure I understand the blocksize thing - why not read the file size first, then create a seq of that size, then read the data into it in a single read? this impl reallocates and copies the contents leading to basically `O(N^2)` overhead

stew/io2.nim
```diff
@@ -0,0 +1,1007 @@
...
+  ## Creates the full path ``path`` with mode ``createMode``.
+  ##
+  ## Path may contain several subfolders that do not exist yet.
+  ## The full path is created. If this fails, error will be returned.
+  ##
+  ## It does **not** fail if the folder already exists because for
+  ## most usages this does not indicate an error.
+  let paths = getPathItems(path, true)
+  when defined(posix):
+    let oldmask = posix.umask(Mode(0))
```
alaviss, line 531: This doesn't seem to be used anywhere meaningful, `rawCreateDir` doesn't modify the umask.

alaviss (commented):
The `Permissions` type seems to be unused. Shouldn't there be a conversion routine between it and numeric permissions, given that the numeric permission is used basically everywhere in this module.

stew/io2.nim
cheatfate: This procedure accepts path as an argument and so it do not supports append operation, it is originally designed to truncate original file and replace information in it. So its why we resetting permissions on file.

stew/io2.nim
cheatfate: Without umask reset its impossible to create directory with all possible permissions combinations.
https://www.cyberciti.biz/tips/understanding-linux-unix-umask-value-usage.html

cheatfate: `rawCreateDir()` is a private procedure, so to avoid changing `umask` for every create directory we set/restore `umask` per path.

alaviss: That seems a bit dangerous, `umask` change the mask of the entire process. Seems like a race waiting to happen in MT programs.

> Without umask reset its impossible to create directory with all possible permissions combinations.

Why does this matter again? This module doesn't seem to care about that part for files (in fact it clearly clarify how it interacts with umask). Isn't the point of umask is for the user to have total control over the resulting permissions?

arnetheduck (approved):

stew/io2.nim
```diff
@@ -0,0 +1,1308 @@
...
+  ## It does **not** fail if the folder already exists because for
+  ## most usages this does not indicate an error.
+  let paths = getPathItems(path, true)
+  for item in paths:
+    let res = rawCreateDir(item, createMode)
+    if res.isErr():
+      return err(res.error)
+  ok()
+
+proc toSet*(mask: int): Permissions =
```
alaviss, line 616: This is a terrible name for a proc that acts on a primitive type like `int`. Maybe `toPermissions` would work better?

PR #55, opened 2020-09-30
Console input/output procedures.

cheatfate:
Different console procedures which are missing or not properly implemented in Nim' stdlib.
Current PR introduces cross-platform password reading procedure which:
1. Supports reading of UTF-8 passwords. (Does not supported on Nim 1.2.6 stdlib).
2. Supports reading from both console and redirected stdin. (Does not supported on Nim devel stdlib).
3. Supports password size limitation (Removes attack vector which is present in Nim 1.2.6/devel stdlib).
4. Do not raise exceptions (Uses io2.IoResult[T]).

### DIFF
--- a/stew/conio.nim
+++ b/stew/conio.nim
@@ -305,8 +305,9 @@ elif defined(posix):
         let res = read(STDIN_FILENO, cast[pointer](addr buffer[0]),
                        len(buffer))
         if res < 0:
+          let errCode = ioLastError()
           discard tcSetAttr(STDIN_FILENO, TCSADRAIN, addr(old))
-          return err(ioLastError())
+          return err(errCode)
 
         if tcSetAttr(STDIN_FILENO, TCSADRAIN, addr(old)) != cint(0):
           return err(ioLastError())
--- a/stew/conio.nim
+++ b/stew/conio.nim
@@ -7,7 +7,7 @@
 ## those terms.
 
 ## This module implements cross-platform console procedures.
-import io2
+import io2, utf8
 export io2
 
 when defined(windows):
@@ -62,6 +62,7 @@ when defined(windows):
     ENABLE_PROCESSED_INPUT = 0x0001'u32
     ENABLE_ECHO_INPUT = 0x0004'u32
     FILE_TYPE_CHAR = 0x0002'u32
+    ERROR_NO_UNICODE_TRANSLATION = 1113'u32
 
   proc isConsoleRedirected*(hConsole: uint): bool =
     ## Returns ``true`` if console handle was redirected.
@@ -73,7 +74,7 @@ when defined(windows):
     else:
       true
 
-  proc readConsoleInput(maxBytes: int): IoResult[string] =
+  proc readConsoleInput(maxChars: int): IoResult[string] =
     let hConsoleInput =
       block:
         let res = getStdHandle(STD_INPUT_HANDLE)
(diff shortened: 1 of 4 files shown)


stew/conio.nim
```diff
@@ -0,0 +1,337 @@
...
+        return err(errCode)
+
+      if setConsoleOutputCP(prevOutputCP) == 0'i32:
+        return err(ioLastError())
+    ok()
+
+elif defined(posix):
+  import posix, termios
+
+  proc isConsoleRedirected(consoleFd: cint): bool =
```
zah, line 260: In Chronicles, I too need to check if `stdin` is redirected. Let's export something that would allow me perform this check in cross-platform way.

stew/conio.nim
```diff
@@ -0,0 +1,337 @@
...
+        return err(errCode)
+
+      if setConsoleCP(prevInputCP) == 0'i32:
+        return err(ioLastError())
+
+      # Truncate additional bytes from buffer.
+      buffer.setLen(int(min(bytesRead, uint32(maxBytes))))
+
+      # Trim CR/CRLF from buffer.
+      if len(buffer) > 0:
```
zah, line 115: This looks like duplicating `strutils.stripLineEnd`.

stew/conio.nim
```diff
@@ -0,0 +1,337 @@
...
+        return err(errCode)
+
+      # Restore previous console CodePage.
+      if setConsoleCP(prevInputCP) == 0'i32:
+        return err(ioLastError())
+
+      # Truncate additional bytes from buffer.
+      buffer.setLen(int(min(charsRead, uint32(maxBytes))))
+      # Truncate CRLF in result wide string.
+      if len(buffer) > 0:
```
zah, line 175: Ditto (`strutils.stripLineEnd`)

cheatfate: This works on `int16` chunks, not on chars/bytes.

stew/conio.nim
cheatfate: The problem of `strutils.stripLineEnd` is that it thinks that `\v` and `\f` are line end, but this is not a proper markers of end of line.

zah: These characters are quite obscure nowadays, but OK - if this is a general purpose procedure for reading a line of characters from the screen, why should it preserve a page break character at the end of the line, but not a new line character in the same place?

cheatfate: @zah we are supposed to read UTF-8 encoded strings are you sure `\v` and `\f` could not be part of UTF-8 encoded symbol which was cut because of buffer limits?

zah: Yes, in UTF-8, the first 128 characters in the ASCII table are encoded with a single byte that has a leading zero bit. All other characters are encoded with a sequence of bytes where each byte has a leading bit set to 1. More details here:

https://en.wikipedia.org/wiki/UTF-8#Encoding

cheatfate: @zah  you are right, but i still think `\v` and `\f` is not a proper condition for line end which we are going to receive from console. When we reading from console the only proper line end is `\n` and `\r\n` and nothing else, `strutils` procedure is incorrect.

jangko:
I think we can extract the utf8 module + tests and combine it with my utf-8/16 module. we have some overlap functionality and some are distinct.

stew/utf8.nim
```diff
@@ -0,0 +1,466 @@
...
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+## This module implements UTF-8 related procedures.
+import results, io2
+export results
+
+type
+  UResult*[T] = Result[T, cstring]
+  Wides32* = int32 | uint32
```
arnetheduck, line 15: `std/unicode` and `Rune`? this way, it would integrate with other nim libraries

stew/utf8.nim
```diff
@@ -0,0 +1,466 @@
...
+  Wides16* = int16 | uint16
```
arnetheduck, line 16: unused

stew/utf8.nim
```diff
@@ -0,0 +1,466 @@
...
+  Bytes* = int8 | char | uint8 | byte
+
+const
+  ErrorBufferOverflow* = cstring"Buffer is not large enough"
+  ErrorInvalidSequence* = cstring"Invalid Unicode sequence found"
+  ErrorInvalidLocale* = cstring"Could not obtain system locale"
+  ErrorNotEnoughCharacters* = cstring"Not enough characters in string"
+
+proc utf8Validate*[T: Bytes](data: openarray[T]): bool =
```
arnetheduck, line 25: std/unicde.validateUTF8?

tersec:
Why do the UTF-8/32 parts?

PR #57, opened 2020-10-02
add uint support in endians2 to match beacon_chain/ssz/bitseqs

tersec:
Otherwise, nbc can't (readily, without its own workarounds) switch to using `endians2` instead of its own code. bitseqs uses `uint`, so even though non-size-specified types might not prove ideal for serialization, it's still useful for endians2 to support it.


tersec:
Closing in favor of alternative approach.

PR #61, opened 2020-11-03
Fix getCurrentDir() gcc compilation warning.

cheatfate:


stefantalpalaru:
If it fixes this issue, we can close it: https://github.com/status-im/nim-stew/issues/60

PR #64, opened 2020-12-09
random helpers

arnetheduck:
* arrayops as a home for small array/openArray utilities
* assign2 - a replacement for genericAssign and assignment operators in
general which in nim are very slow
* use assign2 in a few places to speed things up

### DIFF
--- a/stew/assign2.nim
+++ b/stew/assign2.nim
@@ -1,0 +1,50 @@
+import
+  std/typetraits
+
+{.push raises: [Defect].}
+
+func assign*[T](tgt: var openArray[T], src: openArray[T]) =
+  mixin assign
+
+  if tgt.len != src.len:
+    raiseAssert "Target and source lengths don't match: " &
+      $tgt.len & " vs " & $src.len
+
+  when supportsCopyMem(T):
+    if tgt.len > 0:
+      copyMem(addr tgt[0], unsafeAddr src[0], sizeof(tgt[0]) * tgt.len)
+  else:
+    for i in 0..<tgt.len:
+      assign(tgt[i], src[i])
+
+func assign*[T](tgt: var seq[T], src: openArray[T]) =
+  mixin assign
+  tgt.setLen(src.len)
+
+  assign(tgt.toOpenArray(0, tgt.high), src)
+
+func assign*[T](tgt: var T, src: T) =
+  # The default `genericAssignAux` that gets generated for assignments in nim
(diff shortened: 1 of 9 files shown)


stew/assign2.nim
```diff
@@ -0,0 +1,50 @@
...
+    for t, s in fields(tgt, src):
+      when supportsCopyMem(type s) and sizeof(s) <= sizeof(int) * 2:
+        t = s # Shortcut
+      else:
+        assign(t, s)
+  elif T is seq:
+    assign(tgt, src.toOpenArray(0, src.high))
+  elif T is ref:
+    tgt = src
+  elif compiles(distincBase(T)):
```
zah, line 47: This goes a little bit against the spirit of distinct types which are sometimes introduced specifically to provide different behavior for certain operations (in this case, assignment).

zah (approved):

stew/assign2.nim
arnetheduck: `=` works for `distinct` types out-of-the-box - `assign` is pretty a drop-in replacement that fixes the massive inefficiency in the std lib - unfortunately we can't overload `=` and the c-generated assignments to use it - it really needs an upstream fix at which point this function/module becomes obsolete (https://github.com/nim-lang/Nim/issues/15571)

arnetheduck: ah, you mean that we should try to detect specific `=` overloads for those distinct types? hm.. possible?

arnetheduck: if that's the case, should distinct types with `=` really `supportsCopyMem`?

stew/assign2.nim
```diff
@@ -0,0 +1,58 @@
...
+  std/typetraits,
+  ./shims/macros
+
+{.push raises: [Defect].}
+
+func assignImpl[T](tgt: var openArray[T], src: openArray[T]) =
+  mixin assign
+  when supportsCopyMem(T):
+    if tgt.len > 0:
+      copyMem(addr tgt[0], unsafeAddr src[0], sizeof(tgt[0]) * tgt.len)
```
ringabout, line 11: Is it possible that the memory of `tgt` and `src` overlaps? ref https://github.com/nim-lang/Nim/pull/21661

arnetheduck: that's a good point, I didn't consider overlapping arrays (including self-assignment)

arnetheduck: https://github.com/status-im/nim-stew/pull/182

PR #66, opened 2020-12-11
leb128 + bitops fixes

arnetheduck:
Leb128 is a variable-length encoding for unsigned integers that is used
in a number of contexts - in particular, wasm, dwarf and protobuf.

This is an optimized low-level implementation that unrolls the loop
reading/writing the buffer - it is suitable to use as base for a more
specific API - no memory allocations, no exceptions.

This PR also fixes bitops2 to not raise on certain `uint->int`
conversions, adapting `bitops2` to nim 1.0 conversion rules by using a cast
instead of raising on `uint->int` conversion

### DIFF
--- a/stew/leb128.nim
+++ b/stew/leb128.nim
@@ -1,8 +1,9 @@
-## Low-level little-endian base 128 variable length integer operators, as
+## Low-level little-endian base 128 variable length integer/byte converters, as
 ## described in https://en.wikipedia.org/wiki/LEB128 - up to 64 bits supported.
 ##
-## Used in formats like DWARF, WASM - unsigned variant identical to protobuf/go
-## varint.
+## Used in formats like DWARF, WASM - the encoding is fully compatible with
+## unsigned protobuf, go varint:s and can therefore directly be used for them -
+## signed varint:s can easily be implemented on top.
 ##
 ## This implementation contains low-level primitives suitable for building
 ## more easy-to-use API.
@@ -51,7 +52,7 @@ type
 template write7(next: untyped) =
   # write 7 bits of data
   if v > type(v)(127):
-    result.data[result.len] = cast[byte](v) or 0x80'u8
+    result.data[result.len] = cast[byte](v and type(v)(0xff)) or 0x80'u8
     result.len += 1
     v = v shr 7
     next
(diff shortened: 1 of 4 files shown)


stew/leb128.nim
```diff
@@ -0,0 +1,156 @@
...
+    len*: int8 # >= 1 when holding valid protobuf
+
+  Leb128Error* = enum
+    incomplete = "End marker not found in given stream"
+    overflow = "Value would overflow"
+
+template write7(next: untyped) =
+  # write 7 bits of data
+  if v > type(v)(127):
+    result.data[result.len] = cast[byte](v) or 0x80'u8
```
cheatfate, line 54: It is safer to use construction like this, because sometime you can get compiler error/warning.
```nim
result.data[result.len] = cast[byte](v and type(v)(0xFF)) or 0x80'u8
```

stew/leb128.nim
```diff
@@ -0,0 +1,156 @@
...
+        write7(): # 28
+          write7(): # 35
+            write7(): # 42
+              write7(): # 49
+                write7(): # 56
+                  write7(): # 63
+                    discard
+
+  # high bit not set since v <= 127 at this point!
+  result.data[result.len] = cast[byte](v)
```
cheatfate, line 87:
```nim
result.data[result.len] = cast[byte](v and type(v)(0xFF))
```

stew/leb128.nim
```diff
@@ -0,0 +1,156 @@
...
+      return (I(0), -cast[int8]((shift div 7) + 1))
+
+  res = res or vals
+  if b == valb: # High bit not set
+    return (res, cast[int8]((shift div 7) + 1))
+
+func fromBytes*(
+    I: type SomeUnsignedInt,
+    x: openArray[byte],
+    T: type Leb128): tuple[val: I, len: int8] {.noinit.} =
```
cheatfate, line 111: Does this decoding procedure is able to detect overlong values?
https://github.com/status-im/nim-libp2p/blob/master/tests/testvarint.nim#L382

cheatfate: And how this procedure supports 9-byte length LibP2P varints?

arnetheduck: `leb128` doesn't have those things "natively" but it can easily be added on (which is a bit a point of these primitives):

https://github.com/status-im/nim-libp2p/pull/481/files#diff-c078cd08c868d565974266be04690ecd290f63eafaa5f51cfd4e3b9769437868R140

PR #67, opened 2020-12-22
add github action

jangko:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,211 @@
...
+      - name: Setup environment
+        shell: bash
+        run: echo '${{ github.workspace }}/nim/bin' >> $GITHUB_PATH
+
+      - name: Run nim-stew tests
+        shell: bash
+        working-directory: nim-stew
+        run: |
+          nimble install -y --depsOnly
+          nimble install -y stint
```
arnetheduck, line 186: stew doesn't have dependencies, this can be removed (specifically, it doesn't depend on stint)

jangko: that's right, it's an artifact from other repo, I'll remove them.

PR #71, opened 2021-01-22
Allow comparing Result[void, E]

mratsim:
This allow comparing `Result[void, E]` types.

Also simplify the signature

```Nim
func `==`*[T0, E0, T1, E1](lhs: Result[T0, E0], rhs: Result[T1, E1]): bool {.inline.} =
```

The case where `==` is defined for 2 different types is very rare, unused in our case and it's better to an explicit conversion anyway and people can use a converter otherwise.

This might even improve compilation time by being less taxing on the number of generics to keep track of.

### DIFF
--- a/stew/results.nim
+++ b/stew/results.nim
@@ -442,11 +442,14 @@ template capture*[E: Exception](T: type, someExceptionExpr: ref E): Result[T, re
     ret = R.err(caught)
   ret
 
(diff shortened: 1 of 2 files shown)


arnetheduck:
Oops, missed this PR, but..

> types is very rare

not really - ie for any data type with a `key` inside, you'll want to base the comparison on the key itself and not the whole type - this is common enough that C++ was fixed to allow it for example (for lookups in maps etc).

In general, it's better to provide a specific `void` overload like in the other cases - also because `Result` should as little as possible dictate constraints that might stand in the way of its adoption.

PR #74, opened 2021-01-26
results.`==` with void, Reboot #73 addressing comments

mratsim:


arnetheduck (approved):
could have a test for an err as well just to be safe, but..

mratsim:
Not sure what's happening, CI queues seem clogged. merging

PR #75, opened 2021-01-28
Fix #62

### LINKED ISSUE #62: getCustomPragmaFixed does not support more than one pragma parameter
```nim
from macros import getCustomPragmaVal
from stew/shims/macros as stew_macros import hasCustomPragmaFixed, getCustomPragmaFixed

template dbColumnName*(name: string, primaryKey: bool = false) {.pragma.}
    ## Specifies the database column name for the object property

type MyType = object
  name {.dbColumnName("blah", true).}: string

let myType = MyType(name: "eric")

echo type myType.type.getCustomPragmaFixed("name", dbColumnName)
echo type myType.name.getCustomPragmaVal(dbColumnName)
```

Expected results:
```shell
tuple[name: string, primaryKey: bool]
tuple[name: string, primaryKey: bool]
```

Actual results:
```shell
Error Type
tuple[name: string, primaryKey: bool]
```

Reproduction: 
```shell
git clone https://github.com/emizzle/nim-test-env && cd nim-test-env
make get_custom_pragma
```

gabbhack:
- Supports more than one pragma parameter (returns `tuple`)
- Returns `nil` if there are zero parameters in the pragma (like std)

### DIFF
--- a/stew/shims/macros.nim
+++ b/stew/shims/macros.nim
@@ -212,8 +212,19 @@ proc getPragma(T: NimNode, lookedUpField: string, pragma: NimNode): NimNode =
   error "The type " & $Tresolved & " doesn't have a field named " & lookedUpField
 
 macro getCustomPragmaFixed*(T: type, field: static string, pragma: typed{nkSym}): untyped =
+  result = nil
   let p = getPragma(T, field, pragma)
-  if p != nil and p.len == 2: p[1] else: p
+
+  if p != nil and p.len > 0:
+    if p.len == 2:
+      result = p[1]
+    else:
+      let def = p[0].getImpl[3]
+      result = newTree(nnkPar)
+      for i in 1 ..< def.len:
+        let key = def[i][0]
+        let val = p[i]
+        result.add newTree(nnkExprColonExpr, key, val)
 
 macro hasCustomPragmaFixed*(T: type, field: static string, pragma: typed{nkSym}): untyped =
   newLit(getPragma(T, field, pragma) != nil)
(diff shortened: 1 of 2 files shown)


zah:
`getCustomPragmaFixed` was introduced quite some time ago when `getCustomPragma` from the standard library had a number of issues preventing real-world usage within our libraries. Since then, I've helped the Nim team fix the issues and I don't see any reason for new users to depend on `getCustomPragmaFixed`. It's kept around only because we haven't invested the time to rework the code that was based on it. What is your use case and why can't you rely on `getCustomPragma`? To clarify further, instead of changing the API here, we are likely to just drop `getCustomPragmaFixed` in favor of the standard library.

arnetheduck:
@zah can we provide a compatibility shim so that anyone depending on the version in stew is redirected to the upstream one instead? we can mark it as deprecated or whatever, but if possible, we shouldn't unnecessary break user code

gabbhack:
@zah 
I am writing another library for deserialization, and I ran into the problems already described in https://github.com/nim-lang/RFCs/issues/176 (where I found this library):
- Fail with generics objects (https://github.com/nim-lang/Nim/issues/11923)
- Fail with object variants (https://github.com/nim-lang/Nim/issues/11415)
- Fail with pragmas on case field (https://github.com/nim-lang/Nim/issues/14511)

zah:
I'm willing to accept this patch, but it would be a breaking change for a number of our internal packages that already use `getCustomPragmaFixed`:

https://github.com/search?q=org%3Astatus-im+getCustomPragmaFixed&type=code

Would you agree to take a look at the usages and submit suitable PRs for all projects?

zah:
Actually, I think I misunderstood the description and this is not really a breaking change. Apologies.

PR #76, opened 2021-02-10
results: include error when expecting

arnetheduck:


arnetheduck:
windows having mingw download issues, nothing platform-specific in this pr though

PR #78, opened 2021-03-04
Optimized and exception-less encoding/decoding procedures for decimal integers

cheatfate:


stew/base10.nim
```diff
@@ -0,0 +1,180 @@
...
+  if len(src) == 0:
+    return err("Missing decimal value")
+  var v = T(0)
+  for i in 0 ..< len(src):
+    let d =
+      when A is char:
+        if src[i] >= char(0x30'i8) and src[i] <= char(0x39'i8):
+          int(int8(src[i]) - 0x30'i8)
+        else:
+          -1
```
arnetheduck, line 38: why not return err here?

stew/base10.nim
```diff
@@ -0,0 +1,180 @@
...
+    MaxValue = T(high(T) div 10)
+    MaxNumber = T(high(T) - MaxValue * 10)
+
+  if len(src) == 0:
+    return err("Missing decimal value")
+  var v = T(0)
+  for i in 0 ..< len(src):
+    let d =
+      when A is char:
+        if src[i] >= char(0x30'i8) and src[i] <= char(0x39'i8):
```
arnetheduck, line 35: `let sc = src[i]` to avoid the out-of-bounds checking on every access

arnetheduck: it will likely be optimized away, but no point in making the compiler work when it can easily be avoided

stew/base10.nim
```diff
@@ -0,0 +1,179 @@
...
+
+  if len(src) == 0:
+    return err("Missing decimal value")
+  var v = T(0)
+  for i in 0 ..< len(src):
+    let ch = src[i]
+    let d =
+      when A is char:
+        if ch >= char(0x30'i8) and ch <= char(0x39'i8):
+          int(int8(ch) - 0x30'i8)
```
arnetheduck, line 37: the conversion to int here is no longer needed actually - we can go straight to T - without it, we can also remove the push overflowchecks off, and work with unsigned only. also, `char` is actually unsigned in nim so we can also `let ch = when A is char:  byte(src[i]) else: src[i]` which should further simplify the code

arnetheduck (approved):

arnetheduck (approved):

PR #87, opened 2021-09-09
CI: test with multiple Nim branches

stefantalpalaru:


stefantalpalaru:
Moved the failing Nim branches to a daily CI job.

PR #88, opened 2021-09-24
Introduce a `Lent[T]` shim for the lent type and use it in the `results` module

zah:


arnetheduck:
it's a bit early for this - ie it would make a central piece of the code behave in a different way between 1.2 and 1.6 - when we have both versions in unit tests for all relevant repos, we can also start differentiating between the two versions.

zah:
We can have this disabled by default when it's time to switch to Nim 1.6.
For now, having it enabled facilitates testing Nim 1.6.

stew/results.nim
```diff
@@ -458,21 +461,21 @@ func `==`*[E0, E1](lhs: Result[void, E0], rhs: Result[void, E1]): bool {.inline.
...
     lhs.e == rhs.e
 
-func get*[T: not void, E](self: Result[T, E]): T {.inline.} =
+func get*[T: not void, E](self: Result[T, E]): Lent[T] {.inline.} =
   ## Fetch value of result if set, or raise Defect
   ## Exception bridge mode: raise given Exception instead
   ## See also: Option.get
   assertOk(self)
-  self.v
+  return self.v
```
arnetheduck, line 469: this looks like something to fix in the compiler rather than here, or we're in for a massive refactoring

tersec:
`results` has since been moved to a new repository: https://github.com/arnetheduck/nim-results

PR #90, opened 2021-10-07
Jordan/txpool tools

mjfh:


arnetheduck:
Apart from the RB tree (which probably should go by the name of `SortedTable`, just like `Table` isn't called `HashSet`), this PR introduces several fairly specialized and unique collections that solve some very specific problems, to the point that it almost amount to a new collections library, except that the reuse value is .. low .. similar to how the upstream collections library has obvious gaps and too many highly specific collection types that don't solve a "general" problem.

Given that a "general" problem is not being solved here, we're looking at  a significant maintenance cost, specially given that they at the same time diverge from the existing collections with naming for common operations etc and can't fluidly be used with other, common nim idioms - at the same time, they're not enough for a full collection library because with the new naming and "style", they don't cover the basics that get used in every other application out there - I'm having trouble seeing that we will ever reuse them - so what's the .. larger goal / problem that we're solving here? Are we aiming for a full replacement of the stdlib collections or is this just a one-off that solves a particular problem?

PR #91, opened 2021-11-09
Fix generic sandwich problem

cheatfate:
This fix by @Araq actually fixes all my problems with generic sandwiches.
CC @Araq @zah @arnetheduck

### DIFF
--- a/stew/shims/macros.nim
+++ b/stew/shims/macros.nim
@@ -403,7 +403,7 @@ iterator baseTypes*(exceptionType: NimNode): NimNode =
     typ = objType[1][0]
     yield typ
 
-macro unpackArgs*(callee: typed, args: untyped): untyped =
+macro unpackArgs*(callee: untyped, args: untyped): untyped =
   result = newCall(callee)
   for arg in args:
     let arg = if arg.kind == nnkHiddenStdConv: arg[1]


stefantalpalaru (approved):

stefantalpalaru:
I don't understand how, but it does fix this compilation failure for me: https://ci.status.im/blue/organizations/jenkins/nimbus%2Fnimbus-eth2%2Fmultibranch/detail/PR-3074/1/pipeline

Araq:
Explanation: When you require a `typed` expression an overload set (== "sym choice") is closed which is the opposite of what we need for forwarding of `init` procs. It only happened to work for the 1.2.x line as the 1.2.x compiler is so bad that some unrelated `init` proc matches, the compiler ends up calling the wrong proc, the C compiler doesn't notice and the binary happens to work as the `inits` do little more than `zeroMem` anyway.

zah:
@Araq, this looks like a bug though, because in the places `unpackArgs` is used, the `init` expression is a mixed in symbol, so the expectation is that the typed expression entering the macro is a `nkOpenSymChoice`.

Anyway, I'm glad the issue is resolved. Using `untyped` expressions should be safe.

PR #94, opened 2021-12-02
Sets enhancements

zah:


stew/templateutils.nim
```diff
@@ -0,0 +1,36 @@
...
+  iterator evalTemplateParamOnceImpl[T](x: var T): CppVar[T] =
+    yield CppVar[T](addr(x))
+
+  template stripCppVar[T](p: CppVar[T]): var T =
+    ((ptr T)(p))[]
+else:
+  iterator evalTemplateParamOnceImpl[T](x: var T): var T =
+    yield x
+
+template evalTemplateParamOnce*(templateParam, newName, blk: untyped) =
```
arnetheduck, line 18: what's the advantage over `let x = xParam`?

zah: Two benefits:

* It avoids copying whenever possible.
* It works for var parameters.

arnetheduck: ok - I tested it with:

```
template foo(xParam: untyped) =
  evalTemplateParamOnce(xParam, x):
    echo x
    echo x

foo(@[123] & @[2134])

let v = @[432]
foo(v)

```

and in both cases, there's a temporary + copy involved  (ie also with v when in theory, it's not needed) :/

stew/templateutils.nim
```diff
@@ -0,0 +1,53 @@
...
+  ## * It works for var parameters.
+  ##
+  ## Usage example:
+  ##
+  ## template foo(xParam: SomeType) =
+  ##   evalTemplateParamOnce(xParam, x):
+  ##     echo x
+  ##     echo x
+  ##
+  ##  A currently existing limitation is that the `evalTemplateParamOnce`
```
arnetheduck, line 33: I've found that in most cases this can be worked around by having the template return `untyped` - wouldn't that work here?

ie `template foo(): untyped = "hello"` and `template foo(): untyped = discard` both work

zah: No, the limitation is coming from the fact that the body of the block provided by the user is inserted inside a for loop. It's possible to work-around the limitation by introducing a copy, but since the user code can already do that at the call-site of the template, it would be inappropriate to do in within the template.

arnetheduck: so by and large, this construct sometimes introduces a copy as seen in my example above, is difficult to spell, has complex limitations that are difficult to reason about, and doesn't work with expressions, which is a proven way to make user code safer and easier to reason about in many situations - it doesn't really look like a net win in any case, ie can you proved an example where it does _not_ introduce the copy?

stew/templateutils.nim
zah: The second example doesn't introduce a copy, so I'm not sure how you reached your conclusions. Here is the generated code:

```
	nimln_(12, "/Users/zahary/status/nimbus-eth2/vendor/nim-stew/tests/arne.nim");
	v = (tySequence__qwqHTkRvwhrRyENtudHQ7g*) newSeq((&NTI__qwqHTkRvwhrRyENtudHQ7g_), 1);
	v->data[0] = ((NI) 432);
	{
		{
			tySequence__qwqHTkRvwhrRyENtudHQ7g** paramAddrX60gensym26540001_;
			tyArray__nHXaesL0DJZHyVS07ARPRA T9_;
			tyArray__nHXaesL0DJZHyVS07ARPRA T10_;
			paramAddrX60gensym26540001_ = (tySequence__qwqHTkRvwhrRyENtudHQ7g**)0;
			nimln_(46, "/Users/zahary/status/nimbus-eth2/vendor/nim-stew/stew/templateu"
"tils.nim");
			paramAddrX60gensym26540001_ = (&v);
			nimln_(6, "/Users/zahary/status/nimbus-eth2/vendor/nim-stew/tests/arne.nim");
			nimZeroMem((void*)T9_, sizeof(tyArray__nHXaesL0DJZHyVS07ARPRA));
			T9_[0] = dollar___Gsa4WPnABj84BVdJHxUK1w((*paramAddrX60gensym26540001_));
			echoBinSafe(T9_, 1);
			nimln_(7, "/Users/zahary/status/nimbus-eth2/vendor/nim-stew/tests/arne.nim");
			nimZeroMem((void*)T10_, sizeof(tyArray__nHXaesL0DJZHyVS07ARPRA));
			T10_[0] = dollar___Gsa4WPnABj84BVdJHxUK1w((*paramAddrX60gensym26540001_));
			echoBinSafe(T10_, 1);
		}
	}
```

The key line is `paramAddrX60gensym26540001_ = (&v);` where the address of the input sequence is taken before being printed twice.

zah: My program was a slight modification of yours, mostly so I can more easily look at the generated code:

```nim
import
  ../stew/templateutils

template foo(xParam: untyped) =
  evalTemplateParamOnce(xParam, x):
    echo x
    echo x

proc main =
  foo(@[123] & @[2134])

  let v = @[432]
  foo(v)

main()
```

stew/templateutils.nim
zah: Sorry, I've missed your previous comment. I've added a reply now that demonstrates that no unwanted copies are produced in the provided example. You can also look at the provided test cases which further prove that.

zah: I will also mention that the `keepItif` operation added to the `sets` module is a good example for something enabled by this API. You cannot emulate it with the usual approach of introducing a local variable. There is also another safe way to implement it by introducing a closure, but this would result in much more surprising codegen side-effects and limitations IMO.

stew/templateutils.nim
arnetheduck: after renaming `v` to `vvvvvv` (for clarity, as you say):

```c
    if (vvvvvv__RG7KbPR7Kab6pICXUHR9a9bg) {
      nimGCunrefNoCycle(vvvvvv__RG7KbPR7Kab6pICXUHR9a9bg);
      vvvvvv__RG7KbPR7Kab6pICXUHR9a9bg = NIM_NIL;
    }
    vvvvvv__RG7KbPR7Kab6pICXUHR9a9bg =
        (tySequence__qwqHTkRvwhrRyENtudHQ7g *)newSeqRC1(
            (&NTI__qwqHTkRvwhrRyENtudHQ7g_), 1);
    vvvvvv__RG7KbPR7Kab6pICXUHR9a9bg->data[0] = ((NI)432);
    {
      {
        tyArray__nHXaesL0DJZHyVS07ARPRA T9_;
        tyArray__nHXaesL0DJZHyVS07ARPRA T10_;
        genericSeqAssign(
            (&paramAddrX60gensym26515001___1FRaij9cAw0pBVhEf9bhR0Bw),
            vvvvvv__RG7KbPR7Kab6pICXUHR9a9bg, (&NTI__qwqHTkRvwhrRyENtudHQ7g_));
        nimZeroMem((void *)T9_, sizeof(tyArray__nHXaesL0DJZHyVS07ARPRA));
        T9_[0] = dollar___Gsa4WPnABj84BVdJHxUK1w(
            paramAddrX60gensym26515001___1FRaij9cAw0pBVhEf9bhR0Bw);
        echoBinSafe(T9_, 1);
        nimZeroMem((void *)T10_, sizeof(tyArray__nHXaesL0DJZHyVS07ARPRA));
        T10_[0] = dollar___Gsa4WPnABj84BVdJHxUK1w(
            paramAddrX60gensym26515001___1FRaij9cAw0pBVhEf9bhR0Bw);
        echoBinSafe(T10_, 1);
      }
    }
```


you can see `v` being assigned to `paramAddrX60gensym26515001___1FRaij9cAw0pBVhEf9bhR0Bw` with a `genericSeqAssign`.

I guess it's a difference of top-level vs in-proc then?

stew/shims/hashes.nim
```diff
@@ -0,0 +1,10 @@
+import
+  std/[hashes, typetraits]
+
+func hashAllFields*(x: object|tuple): Hash =
```
arnetheduck, line 4: this could simply be `hashFields` - `all` is kind of implicit (same as the `fields` magic iterator etc)

stew/templateutils.nim
arnetheduck: would https://github.com/status-im/nim-stew/blob/master/stew/results.nim#L662 be a good candidate for using this function?

PR #96, opened 2022-01-01
`Result` refresh

arnetheduck:
* add full support for `Result[T, void]` (aka `Opt[T]` aka `Option[T]`)
* expand tests
* add `flatten`, `filter` of `Option` fame
* add `tryError` that raises a regular exception when result holds a
value
* fix `$` to print `ok`/`err` in lower-case, like the functions that
created the result
* add `orErr` that collapses all errors to a single value of a
potentially different type - useful when translating errors between
layers
* avoid redundant error message when converting error to string
* avoid multiple evaluation in `valueOr`
* add `unsafeError` to match `unsafeGet`
* let `valueOr` evaluate a block
* add `errorOr` to mirror `valueOr`

### DIFF
--- a/stew/results.nim
+++ b/stew/results.nim
@@ -805,6 +805,14 @@ template valueOr*[T: not void, E](self: Result[T, E], def: untyped): T =
   ## Fetch value of result if set, or evaluate `def`
   ## `def` is evaluated lazily, and must be an expression of `T` or exit
   ## the scope (for example using `return` / `raise`)
+  ##
+  ## Example:
+  ## ```
+  ## let
+  ##   v = Result[int, string].err("hello")
+  ##   x = v.valueOr: 42 # x == 42 now
+  ##   y = v.valueOr: raise (ref ValueError)(msg: "v is an error, gasp!")
+  ## ```
   let s = (self) # TODO avoid copy
   if s.o: s.v
   else: def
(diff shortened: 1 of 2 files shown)


stew/results.nim
```diff
@@ -144,16 +162,53 @@ type
     ##   meta-data collection a visible part of your API in another way - this
     ##   way it remains discoverable by the caller!
     ##
-    ## A natural "error API" progression is starting with `Option[T]`, then
+    ## A natural "error API" progression is starting with `Opt[T]`, then
     ## `Result[T, cstring]`, `Result[T, enum]` and `Result[T, object]` in
     ## escalating order of complexity.
     ##
+    ## # Result equivaleces with other types
```
arnetheduck, line 169:
```suggestion
    ## # Result equivalences with other types
```

stew/results.nim
```diff
@@ -542,111 +770,84 @@ func error*[T, E](self: Result[T, E]): E =
...
+  mixin raiseResultOk
+  if self.o: self.raiseResultOk()
+  when E isnot void:
+    self.e
+
+# Alternative spellings for get
+template value*[T, E](self: Result[T, E]): T = self.get()
+template value*[T: not void, E](self: var Result[T, E]): var T = self.get()
+
+template valueOr*[T: not void, E](self: Result[T, E], def: T): T =
```
zah, line 788: Still suffers from double evaluation of `self`

stew/results.nim
```diff
@@ -542,111 +770,84 @@ func error*[T, E](self: Result[T, E]): E =
...
+      err(Result[T, E])
+    else:
+      err(Result[T, E], self.error)
 
-func mapErr*[E, A](
-    self: Result[void, E], f: proc(x: E): A): Result[void, A] {.inline.} =
-  ## Transform error using f, or return value
-  if self.o: result.ok()
-  else: result.err(f(self.e))
+func filter*[T, E](
```
zah, line 804: I personally wouldn't want to see this monadic style handling errors being overused in the project.

arnetheduck: the core motivation here is to allow `Opt` to offer the same API as `Option` - the function definition is a generalization of the same concept

stew/results.nim
arnetheduck: 9bcba3ade870c30d92e4614461380a56f94179a6

PR #106, opened 2022-03-10
prevent leakage of ptrops checks pragma disabling

tersec:
Because of https://github.com/nim-lang/Nim/issues/19603; fixes https://github.com/status-im/nimbus-eth2/issues/3456

### DIFF
--- a/stew/ptrops.nim
+++ b/stew/ptrops.nim
@@ -1,5 +1,5 @@
 # stew
-# Copyright 2018-2019 Status Research & Development GmbH
+# Copyright 2018-2022 Status Research & Development GmbH
 # Licensed under either of
 #
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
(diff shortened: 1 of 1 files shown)


stew/ptrops.nim
```diff
@@ -36,9 +36,12 @@ template offset*[T](p: ptr T, count: int): ptr T =
   # better optimizations.
 
   # We turn off checking here - too large counts is UB
-  {.checks: off.}
-  let bytes = count * sizeof(T)
-  cast[ptr T](offset(cast[pointer](p), bytes))
+  {.push checks: off.}
```
arnetheduck, line 39: the other option is that we cast things to `uint` instead of messing with checks

tersec: That would also avoid the issue it's triggering in 1.2 with some macro expansion error message in CI.

tersec:
Oh, the CI failures aren't new to this commit:
2022-03-11-100107_524x351_scrot
displays the exact same pattern.

stew/ptrops.nim
tersec: Ah, see elsewhere, those CI failures were already there.

tersec: For the purposes of this PR, I think I want to keep it very minimal -- removing `checks` by using `uint` is laudable, but slightly riskier. This is a fix we can deploy immediately, because it just matches the actual execution model to the intended execution model on the callers' side.

PR #107, opened 2022-03-11
remove ptrops generic distance()

tersec:
Is there a reliable way to find out whether anyone in Status uses this? `nimbus-eth2` doesn't appear to. Haven't tried `nimbus-eth1` yet.

The whole thing is marked as deprecated, so might be safe to remove.

It gets rid of one of the `{.checks: off.}` (albeit, not the one directly responsible for https://github.com/status-im/nimbus-eth2/issues/3456, but lying in wait all the same) altogether. Better and less risky than fixing code no one might be using in a deprecated module.

### DIFF
--- a/stew/ptrops.nim
+++ b/stew/ptrops.nim
@@ -1,5 +1,5 @@
 # stew
-# Copyright 2018-2019 Status Research & Development GmbH
+# Copyright 2018-2022 Status Research & Development GmbH
 # Licensed under either of
 #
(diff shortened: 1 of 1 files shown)


arnetheduck:
> The whole thing is marked as deprecated, 

wait what? where?

tersec:
> > The whole thing is marked as deprecated,
> 
> wait what? where?

Ah, they haven't been. While the uses of this aren't obvious at compile-time, something goes terribly wrong running the test suite, and since the point of this was to be less messy/avoid investment into something which maybe shouldn't exist, it doesn't make sense to pursue this further.

PR #109, opened 2022-04-14
avoid zeroMem when resizing trivial seq:s

arnetheduck:
might as well grab a new sequence instead

### DIFF
--- a/stew/assign2.nim
+++ b/stew/assign2.nim
@@ -32,7 +32,14 @@ func assign*[T](tgt: var openArray[T], src: openArray[T]) =
 func assign*[T](tgt: var seq[T], src: openArray[T]) =
   mixin assign
 
-  tgt.setLen(src.len)
+  when T is SomeNumber:
+    # `setLen` does costly zero:ing both when growing _and_ shrinking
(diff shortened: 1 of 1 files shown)


stew/assign2.nim
```diff
@@ -29,7 +29,14 @@ func assign*[T](tgt: var openArray[T], src: openArray[T]) =
 func assign*[T](tgt: var seq[T], src: openArray[T]) =
   mixin assign
 
-  tgt.setLen(src.len)
+  when T is SomeNumber:
+    # `setLen` does costly zero:ing both when growing _and_ shrinking
+    if tgt.len > src.len or tgt.len < src.len div 4:
+      tgt = newSeqUninitialized[T](src.len)
```
zah, line 35: Memory allocation and garbage collection is faster than skipping some `memset(0)` calls? Doesn't sounds universally true.

arnetheduck: it's true for large allocations above all because allocation doesn't depend that much on the size of the allocation -  it seemed to give an improvement in vtune - see also https://github.com/nim-lang/Nim/issues/19728 and https://github.com/nim-lang/Nim/issues/19727

PR #113, opened 2022-05-23
Revert "Add tests for recordFields being used with empty objects"

arnetheduck:
This reverts commit b87fd80b0f1a4c3c3961e61fbea177ded22b34e0.

The tests are broken for nim-1.2 - this reverts them pending a PR with an actual fix (to prevent build breakage)


arnetheduck:
cc @zah

zah:
There is no need to revert everything. Only the `ptr object` test is problematic. I'll push a fix.

zah:
Fixed in https://github.com/status-im/nim-stew/pull/114

PR #115, opened 2022-05-24
Add enum utilities and add enum with holes support

lchenut:
Add one macro and one procedure:
- macro enumRangeOrd(e: type[enum]): untyped
Creates an array with all the ordinal value of an enum (sorted by default from what I saw)
```nim
type
  Test = enum
    A1 = 1, B1 = 2, C1 = 3
  TestWithHoles = enum
    A2 = 1, B2 = 3, C2 = 5
assert(enumRangeOrd(Test) == [ 1, 2, 3 ])
assert(enumRangeOrd(TestWithHoles) == [ 1, 3, 5 ])
```
- proc contains(e: type[enum], v: SomeInteger): bool
With the macro described above, it checks if an integer is in the enum, in order to use someInt in / notin RandomEnum

In addition to that, I use those two macro/proc to implement the hole management in the macro hasHoles and the proc checkEnumAssign

### DIFF
--- a/stew/objects.nim
+++ b/stew/objects.nim
@@ -1,5 +1,6 @@
 import
-  macros
+  macros,
+  sequtils
 
 template init*(lvalue: var auto) =
   mixin init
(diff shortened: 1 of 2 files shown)


stew/objects.nim
```diff
@@ -67,41 +68,34 @@ proc baseType*(obj: RootObj): cstring =
...
-    return newLit(true)
+macro enumRangeOrd*(a: type[enum]): untyped =
+  ## This macro returns an array with all the ordinal values of an enum
+  let
+    values = a.getType[1][1..^1]
+    valuesOrded = values.mapIt(newCall("ord", it))
+  newNimNode(nnkBracket).add(valuesOrded)
+
+proc contains*(e: type[enum], v: SomeInteger): bool =
+  v in enumRangeOrd(e)
```
Menduist, line 79: You could optimize for non-holes enum a bit here:
```nim
when e.hasHoles:
  v in enumRangeOrd(e)
else:
  v in e.low ..< e.high
```

lchenut: Done it, but the _high_ must be included though.

Menduist (changes requested):
This PR doesn't work when used with other types than `int`

Eg, `5.byte in SomeEnum`

Also, since `hasHoles` is used in `contains`, put it first

Menduist (approved):

stew/objects.nim
```diff
@@ -67,41 +68,36 @@ proc baseType*(obj: RootObj): cstring =
...
+  let len = T.getType[1].len - 2
+
+  quote: `T`.high.ord - `T`.low.ord != `len`
+
+proc contains*[I: SomeInteger](e: type[enum], v: I): bool =
+  when I is uint64:
+    if v > int.high.uint64:
+      return false
+  when e.hasHoles():
+    v.int64 in enumRangeOrd(e).mapIt(it.int64)
```
zah, line 90: Using `mapIt` here would be a bit inefficient as it would allocate a sequence just to perform the check. It would be better to directly produce an `int64` array in `enumRangeOrd`.

PR #116, opened 2022-06-06
Cross-platform lockFile()/unlockFile() procedures for io2.

cheatfate:


stew/io2.nim
```diff
@@ -1312,3 +1361,112 @@ proc readAllChars*(pathName: string): IoResult[string] =
...
   readAllBytes(pathName)
+
+proc lockFile*(handle: IoHandle, kind: LockType, offset: int,
+               size: int): IoResult[void] =
+  ## Apply shared or exclusive file segment lock for file handle ``handle`` and
+  ## range specified by ``offset`` and ``size`` parameters.
+  ##
+  ## ``kind`` - type of lock (shared or exclusive). Please not that only
+  ## exclusive locks has cross-platform compatible behavior. But exclusive
+  ## locks require ``handle`` to be opened for writing.
```
etan-status, line 1372:
```suggestion
  ## ``kind`` - type of lock (shared or exclusive). Please note that only
  ## exclusive locks have cross-platform compatible behavior. However, 
  ## exclusive locks require ``handle`` to be opened for writing.
```

stew/io2.nim
```diff
@@ -206,6 +230,15 @@ elif defined(posix):
...
       F_NOCACHE = cint(48)
 
+  type
+    FlockStruct* {.importc: "struct flock", final, pure,
+                   header: "<fcntl.h>".} = object
+      ltype* {.importc: "l_type".}: cshort
+      lwhence* {.importc: "l_whence".}: cshort
+      start* {.importc: "l_start".}: int
+      length* {.importc: "l_len".}: int
+      pid* {.importc: "l_pid".}: Pid
```
etan-status, line 240:
```suggestion
      ltype* {.importc: "l_type".}: cshort
      lwhence* {.importc: "l_whence".}: cshort
      start* {.importc: "l_start".}: Off
      length* {.importc: "l_len".}: Off
      pid* {.importc: "l_pid".}: Pid
```

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/fcntl.h.html
start / length are technically an `off_t`

stew/io2.nim
```diff
@@ -206,6 +230,15 @@ elif defined(posix):
       O_CLOEXEC = cint(0x1000000)
       F_NOCACHE = cint(48)
 
+  type
+    FlockStruct* {.importc: "struct flock", final, pure,
```
etan-status, line 234: Nim's posix module already defines this as `Tflock`.

stew/io2.nim
```diff
@@ -1312,3 +1361,112 @@ proc readAllChars*(pathName: string): IoResult[string] =
...
+        let errCode = ioLastError()
+        if errCode == EINTR:
+          continue
+        else:
+          return err(errCode)
+      else:
+        return ok()
+  elif defined(windows):
+    let (lowOffsetPart, highOffsetPart) = makeUint32(uint64(offset))
+    let (lowSizePart, highSizePart) = makeUint32(uint64(size))
```
etan-status, line 1399: These could trigger defect on Nim 1.6 when a negative offset / size is passed. 
On Nim 1.2 it would convert the bit pattern, but Windows would then interpret it as a large positive number.

Return early error if these are negative, or use `uint64` in the args (to match `getFileSize` return type)

stew/io2.nim
```diff
@@ -1312,3 +1361,112 @@ proc readAllChars*(pathName: string): IoResult[string] =
...
+    else:
+      ok()
+
+proc lockFile*(handle: IoHandle, kind: LockType): IoResult[IoLockHandle] =
+  ## Apply exclusive or shared lock to whole file specified by file handle
+  ## ``handle``.
+  ##
+  ## ``kind`` - type of lock (shared or exclusive). Please not that only
+  ## exclusive locks has cross-platform compatible behavior. But exclusive
+  ## locks require ``handle`` to be opened for writing.
```
etan-status, line 1450:
```suggestion
  ## ``kind`` - type of lock (shared or exclusive). Please note that only
  ## exclusive locks have cross-platform compatible behavior. However, 
  ## exclusive locks require ``handle`` to be opened for writing.
```

stew/io2.nim
```diff
@@ -1312,3 +1361,112 @@ proc readAllChars*(pathName: string): IoResult[string] =
...
+  ##
+  ## On success returns ``IoLockHandle`` object which could be used for unlock.
+  let size =
+    block:
+      let res = ? getFileSize(handle)
+      when sizeof(int) == 4:
+        if res > int64(high(int)):
+          high(int)
```
etan-status, line 1458: If `lockFile` is updated to use `Off` or `uint64` as arg, this could be replaced with an assert, as `getFileSize` uses `fstat` which returns an `off_t`, and `lockFile` uses `fcntl` which takes an `off_t`.

stew/io2.nim
```diff
@@ -97,6 +99,13 @@ when defined(windows):
       changeTime: uint64
       fileAttributes: uint32
 
+    OVERLAPPED* {.pure, inheritable.} = object
+      internal*: uint64
+      internalHigh*: uint64
+      offset*: uint32
+      offsetHigh*: uint32
+      hEvent*: IoHandle
```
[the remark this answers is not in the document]
etan-status: Nim's `winlean` already defines `OVERLAPPED`

stew/io2.nim
```diff
@@ -97,6 +99,13 @@ when defined(windows):
       changeTime: uint64
       fileAttributes: uint32
 
+    OVERLAPPED* {.pure, inheritable.} = object
```
etan-status, line 102: Nim's `winlean` already defines `OVERLAPPED`

tests/test_io2.nim
```diff
@@ -521,3 +525,213 @@ suite "OS Input/Output procedures test suite":
...
+        TestResult(output: strip(res3.output), status: res3.exitCode),
+      ])
+
+    proc performTest(): IoResult[void] =
+      let path1 = "testfile.lock"
+
+      when defined(windows):
+        const
+          ERROR_LOCK_VIOLATION = 33
+          ERROR_SHARING_VIOLATION = 32
```
etan-status, line 583: These are also defined in `winlean`

tests/test_helper.nim
```diff
@@ -0,0 +1,41 @@
+import std/[os, strutils]
```
etan-status, line 1:
```suggestion
## Copyright (c) 2022 Status Research & Development GmbH
## Licensed under either of
##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
## at your option.
## This file may not be copied, modified, or distributed except according to
## those terms.

import std/[os, strutils]
```

Other files also need bumping to 2022

etan-status (approved):

stew/io2.nim
```diff
@@ -163,9 +172,19 @@ when defined(windows):
...
                         dwMoveMethod: uint32): int32 {.
        importc: "SetFilePointerEx", dynlib: "kernel32", stdcall.}
+  proc lockFileEx(hFile: uint, dwFlags, dwReserved: uint32,
+                  nNumberOfBytesToLockLow, nNumberOfBytesToLockHigh: uint32,
+                  lpOverlapped: pointer): uint32 {.
+        importc: "LockFileEx", dynlib: "kernel32", stdcall, sideEffect.}
+  proc unlockFileEx(hFile: uint, dwReserved: uint32,
+                    nNumberOfBytesToLockLow, nNumberOfBytesToLockHigh: uint32,
+                    lpOverlapped: pointer): uint32 {.
+        importc: "UnlockFileEx", dynlib: "kernel32", stdcall, sideEffect.}
```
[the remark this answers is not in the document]
cheatfate: According to @Araq all the procedure definitions should be marked as `sideEffect`.

stew/io2.nim
cheatfate: It is actually not true, because:
```c
struct flock
  {
    short int l_type;   /* Type of lock: F_RDLCK, F_WRLCK, or F_UNLCK.  */
    short int l_whence; /* Where `l_start' is relative to (like `lseek').  */
#ifndef __USE_FILE_OFFSET64
    __off_t l_start;    /* Offset where the lock begins.  */
    __off_t l_len;      /* Size of the locked area; zero means until EOF.  */
#else
    __off64_t l_start;  /* Offset where the lock begins.  */
    __off64_t l_len;    /* Size of the locked area; zero means until EOF.  */
#endif
    __pid_t l_pid;      /* Process holding the lock.  */
  };
```

cheatfate: So on 64bit platform `l_start` and `l_len` are `__off64_t` instead of `__off_t`

stew/io2.nim
cheatfate: Structure definition is incorrect and do not support 64bit platforms.

tests/test_io2.nim
cheatfate: ditto

stew/io2.nim
cheatfate: This code protects from Windows overflow, because on Windows filesize is always 64bit even if system is only 32bit.

zah: But why would this overflow on Windows? Since `lockFileEx` also takes 64-bit inputs through the "high word", "low word" mechanism, we can just use `int64` for the offset value and this will be compiled to the correct code even on 32-bit systems, no? (since the C compilers support emulated types such as `__int64`)

cheatfate: `io2.lockFile()` accepts `size` as `int` which could be `int32` and `int64` depends on system architecture. So if we running on Windows 32bit we can't use `getFileSize(handle)` result value which is always `int64` because there will be overflow when i'm going to convert `int32(int64 value)`. 
In Linux/BSD field of `stat` structure called `st_size` has `off_t` type which depends on architecture so it will be equal to `int32` in 32bit system. It means that Posix systems do not support files sizes bigger than high(int32). So in this case we going to convert only `int32(int32)` - so its safe.

etan-status: Shouldn't `io2.lockFile()` take the same type that `getFileSize(handle)` result returns? Otherwise, locking may silently fall back to a partial file lock for large files.

Is there a situation where `getFileSize` can return a value that's unsupported for locking?

stew/io2.nim
```diff
@@ -1315,3 +1368,161 @@ proc readAllChars*(pathName: string): IoResult[string] =
...
+          return err(errCode)
+      else:
+        return ok()
+  elif defined(windows):
+    let (lowOffsetPart, highOffsetPart, lowSizePart, highSizePart) =
+      if offset == 0'i64 and size == 0'i64:
+        # We try to keep cross-platform behavior on Windows. And we can do it
+        # because: Locking a region that goes beyond the current end-of-file
+        # position is not an error.
+        (0'u32, 0'u32, 0xFFFF_FFFF'u32, 0xFFFF_FFFF'u32)
```
etan-status, line 1431: Or `uint32.high` / `not 0'u32` instead of `0xFFFF_FFFF'u32`, ensures that the number of `F` is correct. I'm fine with either.

etan-status (approved):
Thanks. The addition of "entire file locking" support beyond 2GB can be quite useful!

stew/io2.nim
```diff
@@ -1315,3 +1368,161 @@ proc readAllChars*(pathName: string): IoResult[string] =
...
+      case kind
+      of LockType.Shared:
+        cshort(posix.F_RDLCK)
+      of LockType.Exclusive:
+        cshort(posix.F_WRLCK)
+    var flockObj =
+      when sizeof(int) == 8:
+        # There is no need to perform overflow check, so we just cast.
+        FlockStruct(ltype: ltype, lwhence: cshort(posix.SEEK_SET),
+                    start: cast[int](offset), length: cast[int](size))
```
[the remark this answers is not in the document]
cheatfate: `when` is compile time check, it means that nim compiler will insert one branch or another, there will be no duplicate.

stew/io2.nim
cheatfate: I'm personally do not like both `high` and `not` because its much more easier to read `0xFFFF_FFFF'u32` for anybody who are not familiar with `nim` constructs.

PR #118, opened 2022-06-15
move `-d:nimRawSetjmp` to config.nim

narimiran:
Refs https://github.com/status-im/nimbus-build-system/issues/44

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,10 +149,6 @@ jobs:
 
       - name: Run tests
         run: |
-          if [[ "${{ matrix.target.os }}" == "windows" ]]; then
-            # https://github.com/status-im/nimbus-eth2/issues/3121
-            export NIMFLAGS="-d:nimRawSetjmp"
-          fi
           nim --version
           nimble --version
           env TEST_LANG=c nimble test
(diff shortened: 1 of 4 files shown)


Menduist:
I've cancelled this run since it will fail on windows and our CI is slightly overloaded :)

narimiran:
> our CI is slightly overloaded

Ok, I'll delay pushing similar changes to other repos for tomorrow.

arnetheduck:
>  (this is only needed in this repo since we test with vcc)

I think better add "not vcc" in all repos - this is not something to research again

arnetheduck:
per discussion in #nim-opportunities, let's use `.cfg` files for simple options

narimiran:
> per discussion in #nim-opportunities, let's use `.cfg` files for simple options

Ok, done. (And I'll make the same change in other repos too)

arnetheduck (approved):

arnetheduck:
So, this looks like it might need investigation on the nim side:


nim.cfg:
```
@if vcc
  -d:testit
@end
```

test.nim:
```
when defined(testit):
  {.error: "vcc in config".}
else:
  {.error: "vcc not in config".}
```

```
nim c --cc:vcc -c test.nim
```

this prints "vcc not in config" on my (linux) nim

narimiran:
> `nim c --cc:vcc -r test.nim`

It seems that `--cc:...` is not picked up at all!
Because if I change the `nim.cfg` to `@if gcc: ...` it says `Error: vcc in config` regardless of what I pass to `--cc`.

PR #123, opened 2022-07-07
leb128: avoid shifts larger than type

arnetheduck:
this may lead to UB

### DIFF
--- a/stew/leb128.nim
+++ b/stew/leb128.nim
@@ -113,7 +113,7 @@ template read7(shift: untyped) =
     b = x[shift div 7]
     valb = b and 0x7f'u8 # byte without high bit
     val = I(valb)
-    vals = val shl shift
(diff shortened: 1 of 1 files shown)


stew/leb128.nim
```diff
@@ -113,7 +113,7 @@ template read7(shift: untyped) =
     b = x[shift div 7]
     valb = b and 0x7f'u8 # byte without high bit
     val = I(valb)
```
mratsim, line 115: This is unsigned

mratsim (commented):
There is no undefined behavior in C when the LHS is unsigned and the RHS is positive or zero, which is the case here
image

And C++ has the same requirement.

arnetheduck:
image

The full paragraph says that if RHS is wider than the size of the integer, the behavior is undefined regardless of sign.

PR #124, opened 2022-07-07
Splitting `all_tests` wrapper

mjfh:
Running some unit tests in a separate file attempts to mitigate the global symbols overflow on Github ci when compiling with nim
  version <= 1.2.


mjfh:
ok, I see.

I guess unittest2 was avoided in order to reduce the imported packages? Otherwise i would prefer to use unittest2 and wrap in the main function wrapper.

arnetheduck:
I'm guessing it hasn't been needed until now - @zah started using it in https://github.com/status-im/nim-stew/pull/122 (stew predates unittest2) - when migrating, it's best to do all unittests in one go so as to avoid issues - it's mostly doable with a search/replace of the import

mjfh:
Well, then I prefer to do it now. I see that is the most clean solution.

PR #125, opened 2022-07-08
Migrate to `unittest2`

mjfh:
Global symbol overflow when running NIM 1.2 on Github ci suggests that unit tests run sort of separately. The replacement library `unittest2`  provides that.


arnetheduck:
Nice!

PR #127, opened 2022-07-18
stylecheck: nnkArglist -> nnkArgList

tersec:
https://github.com/nim-lang/Nim/pull/17529
https://github.com/nim-lang/Nim/pull/19822

PR #128, opened 2022-07-18
fix warnings encountered with Nim 1.6

narimiran:


stew/ranges/stackarrays.nim
```diff
@@ -49,8 +49,11 @@ when defined(windows):
 else:
   proc alloca(n: int): pointer {.importc, header: "<alloca.h>".}
 
+when not declared(RangeDefect):
```
arnetheduck, line 52: it would make sense to create  `stew/shims/stddefects.nim` which creates aliases for all the 1.6-renamed defects

narimiran: Ok, done.

narimiran:
The failed CIs are the same as here: https://github.com/status-im/nim-stew/pull/127#issuecomment-1187079293

PR #34, opened 2020-04-28
wip: general-purpose conversion

arnetheduck:

PR #134, opened 2022-07-22
result: add result assignment extenstion point for `?`

arnetheduck:
When `?` exits early, it does so using variations `return v` - however,
in frameworks such as `chronos`, we need to change `return v` to
`future.complete(v)` - this feature allows chronos to inject a template
that performs this assignment instead of using the "ordinary" flow.

Risk:

`assignResult` might already be taken as a name, and multiple frameworks
might compete for the functionality.

Potential alternative:

Instead of `assignResult`, this construct could be called
`returnResult`, and it would be responsible for breaking the control
flow (performing the `return` in addition to assigning the result).

Other interactions:

https://github.com/status-im/nim-stew/pull/34 proposes a general-purpose
conversion framework - this could be used to automatically convert the
error based on existing conversions - the use case is that oftentimes,
one error needs to be converted to another in a call chain and injecting
a converter simplifies this flow - this might however create an
problematic interaction with an `assignResult` template in the future.

### DIFF
--- a/stew/results.nim
+++ b/stew/results.nim
@@ -1065,18 +1065,44 @@ template `?`*[T, E](self: Result[T, E]): auto =
   ## let v = ? funcWithResult()
   ## echo v # prints value, not Result!
   ## ```
+  ##
+  ## ``assignResult?`` extension point
+  ##
+  ## If a template / function named ``assignResult?`` exists in the scope, it will
+  ## be called instead of assigning the result - this can be used to intercept
+  ## the assignment to `result` that implicitly happens on early return in case
+  ## of error.
+  ##
+  ## To prevent conflicts, it is recommended that ``assignResult?`` is declared
+  ## close to the scope of `?` (as opposed to a globally visible symbol)
+  ##
+  ## ```nim
+  ## proc f(): Result[...] =
+  ##   template `assignResult?`(v: Result) = ...
+  ##   let a = ? f2()
+  ## ```
+  ##
   ## Experimental
   # TODO the v copy is here to prevent multiple evaluations of self - could
   #      probably avoid it with some fancy macro magic..
   let v = (self)
   if not v.oResultPrivate:
-    when typeof(result) is typeof(v):
-      return v
-    else:
-      when E is void:
-        return err(typeof(result))
+    when compiles(`assignResult?`(default(typeof(result)))):
+      when typeof(result) is typeof(v):
+        `assignResult?`(v)
+      elif E is void:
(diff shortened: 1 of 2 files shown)


stew/results.nim
```diff
@@ -953,25 +953,52 @@ template isNone*(o: Opt): bool =
...
   # TODO the v copy is here to prevent multiple evaluations of self - could
   #      probably avoid it with some fancy macro magic..
   let v = (self)
   if not v.o:
-    when typeof(result) is typeof(v):
-      return v
-    else:
-      when E is void:
-        return err(typeof(result))
+    when declared(assignResult):
```
zah, line 986: I'm not sure this `declared` check would actually work for the Chronos use-case. Have you tested it?

stew/results.nim
```diff
@@ -953,25 +953,52 @@ template isNone*(o: Opt): bool =
 # Syntactic convenience
 
 template `?`*[T, E](self: Result[T, E]): auto =
+  mixin assignResult
```
zah, line 956: Since this is a very low-level integration mechanism between advanced Nim libraries, I suggest relying on the Nim `backticks` feature and use an identifier that's extremely unlike to appear in user code (e.g. `#assignResult`). It's also worth re-reading the discussion we had in the Chronos tracker:

https://github.com/status-im/nim-chronos/pull/297#discussion_r927499892

stew/results.nim
arnetheduck: well, there's a test in https://github.com/status-im/nim-chronos/pull/297 - not sure about more complex cases, anything in particular you're worried about?

zah: Ah, OK, I didn't realize the Chronos PR was tested precisely with this version of `Result` (there was the possibility that `assignResult` existed a globally defined template that is overloaded in Chronos).

It would be good to test additional use cases such as generic async procs or async closures defined in generic functions as these trigger substantially different symbol lookup paths in the compiler (which may trip the `declared` check here).

stew/results.nim
arnetheduck: `#assignResult` doesn't compile - `assignResult?` is fine however, switched

stew/results.nim
arnetheduck: added a generic test which seems to work for now - needed an extra mixin _inside_ `assignResult` to not complain about the `result` symbol

arnetheduck:
marking draft for now - failing in 2.0

arnetheduck:
obsoleted by https://github.com/arnetheduck/nim-results/pull/37

PR #140, opened 2022-09-12
fix fromBytes call resolution

arnetheduck:


arnetheduck:
https://github.com/status-im/nim-protobuf-serialization/runs/8303036832?check_suite_focus=true#step:11:144

arnetheduck:
CI unrelated / temporary actions issues

PR #141, opened 2022-09-21
results.Opt as object fields could generate codegen issues.

cheatfate:


arnetheduck:
worked around in https://github.com/status-im/nim-stew/pull/150

PR #144, opened 2022-09-23
Fix assertion error

tavurth:
There was an assertion error being thrown instead of a ValueError

### DIFF
--- a/stew/byteutils.nim
+++ b/stew/byteutils.nim
@@ -46,9 +46,6 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
-  if hexStr.len - sIdx < 1:
-    raise (ref ValueError)(msg: "hex value too short")
-
   # Ensure we have at least some output buffer
   if output.len > 0:
     doAssert(fromIdx >= 0 and toIdx >= fromIdx)
--- a/stew/byteutils.nim
+++ b/stew/byteutils.nim
@@ -49,7 +49,10 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   if hexStr.len - sIdx < 1:
     raise (ref ValueError)(msg: "hex value too short")
 
-  doAssert(fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len)
+  # Ensure we have at least some output buffer
+  if output.len > 0:
+    doAssert(fromIdx >= 0 and toIdx >= fromIdx)
+    doAssert(fromIdx < output.len and toIdx < output.len)
 
   let sz = toIdx - fromIdx + 1
 
--- a/stew/byteutils.nim
+++ b/stew/byteutils.nim
@@ -46,8 +46,10 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
-  if fromIdx < 0 or toIdx < fromIdx or fromIdx >= output.len or toIdx >= output.len:
+  if output.len < 1:
     raise (ref ValueError)(msg: "hex value is invalid")
+
+  doAssert(fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len)
 
   let sz = toIdx - fromIdx + 1
 
--- a/stew/byteutils.nim
+++ b/stew/byteutils.nim
@@ -46,7 +46,7 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
-  if fromIdx <= 0 or toIdx <= fromIdx or fromIdx > output.len or toIdx > output.len:
+  if fromIdx < 0 or toIdx < fromIdx or fromIdx > output.len or toIdx > output.len:
     raise (ref ValueError)(msg: "hex value is invalid")
 
   let sz = toIdx - fromIdx + 1
--- a/stew/byteutils.nim
+++ b/stew/byteutils.nim
@@ -46,7 +46,9 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
-  doAssert(fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len)
+  if fromIdx <= 0 or toIdx <= fromIdx or fromIdx > output.len or toIdx > output.len:
+    raise (ref ValueError)(msg: "hex value is invalid")
+
   let sz = toIdx - fromIdx + 1
 
   if hexStr.len - sIdx < 2*sz:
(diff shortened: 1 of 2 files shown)


stew/byteutils.nim
```diff
@@ -46,7 +46,9 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
-  doAssert(fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len)
+  if fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len:
```
arnetheduck, line 49: we should only raise `ValueError` when `hexStr` is "invalid" - we should assert when `fromIdx`/`toIdx` are inconsistent / out-of-range with respect to `hexStr`

arnetheduck: also, a change like this should be accompanied by a test case

tavurth: Yep was just in the draft of this PR, sorry for notification

tests/test_byteutils.nim
```diff
@@ -65,6 +65,9 @@ suite "Byte utils":
     expect(ValueError): discard hexToByteArray[1]("")
     expect(ValueError): discard hexToByteArray[1]("1")
 
+  test "hexToByteArray: Nully value":
+    expect(ValueError): discard hexToByteArray[1]("0x")
```
tavurth, line 69: @arnetheduck actually I see the line below ([87](https://github.com/status-im/nim-stew/pull/144/files#diff-c456240aa7918dfdb6accf51ef55fd6032b80fa58d2bc699614ff53d505e24cbR87))

When we see a `0x` we return an empty array. Since we are extracting a single value here ValueError seems correct, but just so you can confirm.

arnetheduck: the case where "0x" is given for a 1-byte array, `ValueError` should indeed happen - however, when the caller passes in nonsensical indices, an assertion should happen - your change makes everything a value error which is not the original intent of the function

arnetheduck: keep in mind also that to/from indices are inclusive of their ends in Nim - an "empty" range is thus `0/-1` while `0/0` is a 1-byte range

stew/byteutils.nim
```diff
@@ -46,7 +46,11 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
+  if output.len < 1:
```
arnetheduck, line 49: why are you checking the output length here?

also, even when using the original code without this check, your new test passes, so this change effectively does not make a difference.

tavurth: Oh I'm sorry I made this change in haste!

tavurth: I had a problem because I was on my `feature/execution_api_spec` branch which was checked out while there was error handling in my `web3` impl.

tavurth: How does it look now?

arnetheduck: well, as I mentioned above, your new test case triggers a value error already without any code changes, so the bug you're looking for is elsewhere - this function works as expected, when given "0x" as string

stew/byteutils.nim
```diff
@@ -46,11 +46,16 @@ func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fr
   ## Allows specifying the byte range to process into the array
   var sIdx = skip0xPrefix(hexStr)
 
+  if hexStr.len - sIdx < 1:
```
arnetheduck, line 49: this cannot happen simply because skip0xPrefix is already checking the length the same way - the code will never trigger - if you remove this change, your test case still correctly runs

tavurth: Hmm, in my case removing this does cause my tests in the `nim-web3` repo to fail 🤔 

Seems that it's my use of `openArray` rather than the fixed `byte` length which is causing the issue. I will add more tests for this.

tests/test_byteutils.nim
```diff
@@ -65,6 +67,17 @@ suite "Byte utils":
...
     expect(ValueError): discard hexToByteArray[1]("1")
 
+  test "hexToByteArray: Invalid value":
+    expect(ValueError):
+      var buffer: array[1, byte]
+      hexToByteArray("0x", buffer)
+
+    expect(ValueError):
+      var buffer: distinct seq[byte]
+      hexToByteArray("0x", (openArray[byte])buffer)
```
tavurth, line 77: So I think the problem is here. `openArray[byte].len` is `0` and `openArray.high` returns `-1` I think from your other comment.

It's a little hard to test exactly what it is without transforming everything from `func` to `proc` so I can stick some debugging in there. 🤔

tersec: https://nim-lang.org/docs/system.html#debugEcho%2Cvarargs%5Btyped%2C%5D

arnetheduck: > So I think the problem is here

yes, this would be a problem - in your test, the output buffer is empty, so there's nothing to write to - when an empty output buffer is passed in, we should probably allow this and not raise anything (no valueerror and no assertion)

tavurth: @arnetheduck However in this case we have an invalid hex input `0x`. It's the only case I've found where this bug occurs but I can check some negative cases also.

@tersec 🔥 Thank you!

arnetheduck: > However in this case we have an invalid hex input 0x

I'd consider this a valid serialization of an empty byte sequence with `0x` prefix - ie the caller passed in a 0-byte buffer, so they expect that the hex string is a valid zero-byte representation of the empty buffer, which means either an empty string or just a `0x` prefix

tavurth: @arnetheduck since `openArray` cannot be resized, this seems like it should be some kind of error 🤔

zah, line 77: This line doesn't seem well-formed. It seems a bug to me that you didn't get a compiler error.

If you wan to convert a distinct `seq` type to an open array, the idiomatic code would be to use the `distinctBase` operator and to let the compiler perform an implicit conversion to `openArray`.

tavurth: Ok sure, this is from C++ style days, I've seen it around a few nim repos

```nim
openArray[byte](buffer)
(openArray[byte]) buffer
```
Are all valid. I will switch to `distinctBase`

zah: There are two flavours of `hexStringToArray` functions in this module. The `*Strict` versions require the input string size to match exactly the size of the the output array, while the non-strict versions require only that the input string has enough characters to fill the output array. Both don't resize the output array (as this is not possible with `openArray` as you pointed out yourself).

@arnetheduck's point is that if the user asked for an empty string, it's fine if the non-strict flavour accepts that without an error. The strict flavour is likely to error out unless the input string is also empty (i.e. "" or "0x").

tavurth: Talked with @zah in discord and cleared the use of this as I would like it to be very clear what the problem is here.

tavurth: @zah it seems though like I can ask for an actual value to be filled into a mistakenly empty array:

```nim
  test "valid hex with empty array":
    var buffer: seq[byte]
    hexToByteArray("0x123", openArray[char](buffer))
```

Which will end up giving the user a weird and hard to find bug in their code. Should this not be an assertion error?

stew/byteutils.nim
zah: I also think this should be removed, per the arguments regarding empty output arrays in the discussion below.

PR #146, opened 2022-10-24
Fix compilation on devel

Menduist:
tests are not passing since shallow copies are not a thing in orc. Not sure how to go forward about this, could just disable the shallow tests in devel, but maybe users are relying on this functionality?

### DIFF
--- a/stew/ranges/typedranges.nim
+++ b/stew/ranges/typedranges.nim
@@ -24,22 +24,16 @@ type
   MutByteRange* = MutRange[byte]
 
 proc isLiteral[T](s: seq[T]): bool {.inline.} =
-  when defined(gcOrc) or defined(gcArc):
-    false
-  else:
-    type
-      SeqHeader = object
-        length, reserved: int
-    (cast[ptr SeqHeader](s).reserved and (1 shl (sizeof(int) * 8 - 2))) != 0
+  type
+    SeqHeader = object
+      length, reserved: int
+  (cast[ptr SeqHeader](s).reserved and (1 shl (sizeof(int) * 8 - 2))) != 0
 
 proc toImmutableRange[T](a: seq[T]): Range[T] =
   if a.len != 0:
     when rangesGCHoldEnabled:
-      when declared(shallowCopy):
-        if not isLiteral(a):
-          shallowCopy(result.gcHold, a)
-        else:
-          result.gcHold = a
+      if not isLiteral(a):
+        shallowCopy(result.gcHold, a)
       else:
         result.gcHold = a
     result.start = addr result.gcHold[0]
@@ -141,10 +135,7 @@ proc sliceNormalized[T](r: Range[T], ibegin, iend: int): Range[T] =
                             # an empty range
 
   when rangesGCHoldEnabled:
-    when declared(shallowCopy):
-      shallowCopy(result.gcHold, r.gcHold)
-    else:
-      result.gcHold = r.gcHold
+    shallowCopy(result.gcHold, r.gcHold)
   result.start = r.start.offset(ibegin)
   result.mLen = iend - ibegin + 1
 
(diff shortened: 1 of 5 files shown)


stew/results.nim
```diff
@@ -846,6 +852,8 @@ func flatten*[T, E](self: Result[Result[T, E], E]): Result[T, E] =
     else:
       err(Result[T, E], self.error)
 
+{.push effectsOf: callback.}
```
arnetheduck, line 855: this goes in the function declaration itself so that it's visible - it's critical information for the claler - that said, it also doesn't exist/work on 1.2 so we can't use it there yet

stew/varints.nim
```diff
@@ -113,11 +113,12 @@ func feedByte*(p: var VarintParser, b: byte): VarintState =
   ##     has been exceed. The supplied input can be considered invalid.
   ##
   const maxShift = maxBits type(p)
+  type IntType = p.IntType
```
arnetheduck, line 116: this is some new bug workaround?

stew/ranges/typedranges.nim
```diff
@@ -24,16 +24,22 @@ type
...
 proc toImmutableRange[T](a: seq[T]): Range[T] =
   if a.len != 0:
     when rangesGCHoldEnabled:
-      if not isLiteral(a):
-        shallowCopy(result.gcHold, a)
+      when declared(shallowCopy):
+        if not isLiteral(a):
+          shallowCopy(result.gcHold, a)
+        else:
+          result.gcHold = a
```
arnetheduck, line 42: this defeats the purpose of the ranges entirely - I'd probably go for deprecating the whole library since there's some work on view types being done in the language, but @zah might have some plans?

stew/varints.nim
Menduist: Yup

stew/results.nim
Menduist: https://github.com/status-im/nim-stew/pull/146/commits/738b75f0414920bb8351fe63b6267ad5d9e73ff2

stew/varints.nim
Menduist: dcdbaaeaf3d3d10ac6f0dda4911f0f42e0675a5a
https://github.com/nim-lang/Nim/issues/20645

stew/ranges/typedranges.nim
Menduist: Can we merge this & move this into an issue? Or I can revert the changes to typedranges only fix results, but would be nice to be able to use results with devel

Menduist:
Disabled ranges with `--gc:orc`. There is another failure in the tests of Objects with some flag, but at least everything compiles

deech:
Reported on Discord, the following test doesn't compile with the HEAD of `Nim`:
```nim
suite "Results":
  test "pointer to RootRef":
    type
      O = object of RootRef
        payload:pointer

    proc makeO(p:pointer):Opt[O] =
      if p.isNil:
        result.err()
      else:
        result.ok(O(payload:p))

    check:
      makeO(nil).isNone()
```

it errors with:
```
/home/deech/.cache/nim/all_tests_d/@mtest_results.nim.c:1395:108: error: ‘union <anonymous>’ has no member named ‘_o_1’; did you mean ‘_o_2’?
 1395 | static NIM_CONST tyObject_Result__2ZCWlboZcb9ceT9bFL9amTQ3A TM__k8zgv7NLULF1SW5CdpYwhw_422 = {NIM_FALSE, {._o_1 = {}}}
      |                                                                                                            ^~~~
      |                                                                                                            _o_2
Error: execution of an external compiler program 'gcc -c  -w -fmax-errors=3   -I/home/deech/Nim/lib -I/home/deech/NimWorkspace/nim-stew/tests -o /home/deech/.cache/nim/all_tests_d/@mtest_results.nim.c.o /home/deech/.cache/nim/all_tests_d/@mtest_results.nim.c' failed with exit code: 1
```

I'm not sure if this an error in `Nim` itself or `nim-stew`.

PR #147, opened 2022-11-03
chore: fix 'there should not be `high(value)`. use `high(type)`' deprecation

LNSD:
In _nwaku_ we have been using nim 1.6 for some time. This fixes the following deprecation warning:

```
/nwaku/vendor/nim-stew/stew/sorted_set/rbtree_desc.nim(197, 30) Warning: Deprecated since v1.4; there should not be `high(value)`. Use `high(type)`.; high is deprecated [Deprecated]
```


LNSD:
I see all the _Nim devel_ failing, but the changes have nothing to do with the failing test cases. Is there any reason for that?

cc @zah @arnetheduck

Menduist (approved):

PR #148, opened 2022-11-09
Deprecate baseType

Menduist:
From a quick sweep in our repos, no one uses it
It's based on quite unsafe behavior, doesn't give the same results with arc/orc, and I'm pretty sure a macro could do it, if we need it

### DIFF
--- a/stew/objects.nim
+++ b/stew/objects.nim
@@ -59,15 +59,15 @@ when not compiles(len((1, 2))):
 # Get an object's base type, as a cstring. Ref objects will have an ":ObjectType"
 # suffix.
 # From: https://gist.github.com/stefantalpalaru/82dc71bb547d6f9178b916e3ed5b527d
-proc baseType*(obj: RootObj): cstring {.deprecated.} =
+proc baseType*[T: RootObj](obj: T): cstring {.deprecated.} =
   when not defined(nimTypeNames):
-    raiseAssert("you need to compile this with '-d:nimTypeNames'")
+    {.error: "baseType requires -d:nimTypeNames".}
   when defined(gcArc) or defined(gcOrc):
-    raiseAssert("baseType doesn't work with arc / orc")
+    {.error: "baseType doesn't work with arc / orc".}
   else:
     {.emit: "result = `obj`->m_type->name;".}
 
-proc baseType*(obj: ref RootObj): cstring {.deprecated.} =
+proc baseType*[T: RootObj](obj: ref T): cstring {.deprecated.} =
   obj[].baseType
 
 macro enumRangeInt64*(a: type[enum]): untyped =
--- a/stew/objects.nim
+++ b/stew/objects.nim
@@ -59,13 +59,13 @@ when not compiles(len((1, 2))):
 # Get an object's base type, as a cstring. Ref objects will have an ":ObjectType"
 # suffix.
 # From: https://gist.github.com/stefantalpalaru/82dc71bb547d6f9178b916e3ed5b527d
-proc baseType*(obj: RootObj): cstring =
+proc baseType*(obj: RootObj): cstring {.deprecated.} =
   when not defined(nimTypeNames):
     raiseAssert("you need to compile this with '-d:nimTypeNames'")
   else:
     {.emit: "result = `obj`->m_type->name;".}
 
-proc baseType*(obj: ref RootObj): cstring =
+proc baseType*(obj: ref RootObj): cstring {.deprecated.} =
   obj[].baseType
 
 macro enumRangeInt64*(a: type[enum]): untyped =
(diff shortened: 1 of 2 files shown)


stew/objects.nim
```diff
@@ -59,13 +59,13 @@ when not compiles(len((1, 2))):
 # Get an object's base type, as a cstring. Ref objects will have an ":ObjectType"
 # suffix.
 # From: https://gist.github.com/stefantalpalaru/82dc71bb547d6f9178b916e3ed5b527d
-proc baseType*(obj: RootObj): cstring =
+proc baseType*(obj: RootObj): cstring {.deprecated.} =
```
arnetheduck, line 62: I think for these orc-breaking things, we should add a `when not orc:` in addition to the deprecation so that there's a hard compile error where it's actually broken

ringabout: > I'm pretty sure a macro could do it, if we need it

It is quite unsafe because there is no guarantee regarding the implementation details. Since https://github.com/nim-lang/Nim/pull/20789, the name becomes shorter for ARC/ORC and no longer generates type names for parent objects.

stew/objects.nim
```diff
@@ -59,13 +59,15 @@ when not compiles(len((1, 2))):
...
   when not defined(nimTypeNames):
     raiseAssert("you need to compile this with '-d:nimTypeNames'")
+  when defined(arc) or defined(orc):
+    raiseAssert("baseType doesn't work with arc / orc")
```
arnetheduck, line 66: why not compiletime?

Menduist: It's always compiled since it's not generic

stew/objects.nim
```diff
@@ -59,13 +59,15 @@ when not compiles(len((1, 2))):
 # Get an object's base type, as a cstring. Ref objects will have an ":ObjectType"
 # suffix.
 # From: https://gist.github.com/stefantalpalaru/82dc71bb547d6f9178b916e3ed5b527d
-proc baseType*(obj: RootObj): cstring =
+proc baseType*(obj: RootObj): cstring {.deprecated.} =
   when not defined(nimTypeNames):
     raiseAssert("you need to compile this with '-d:nimTypeNames'")
+  when defined(arc) or defined(orc):
```
ringabout, line 65:
```suggestion
  when defined(gcArc) or defined(gcOrc):
```
there are no defines like `arc` or `orc`.

stew/objects.nim
arnetheduck: well yes, so that's the point: it should fail compilation when any of these are enabled

stew/objects.nim
Menduist: Yeah sorry, thanks

stew/objects.nim
arnetheduck: the whole function should not exist (similar to shallowCopy)

Menduist: Just `nim c stew/objects` fails with `{.error.}` (even without actually using the function)

ringabout:
```nim
when defined(gcArc) or defined(gcOrc):
  proc baseType*(obj: RootObj): cstring {.error: "baseType is not available in ARC/ORC".}
else:
  proc baseType*(obj: RootObj): cstring =
```
Works for me.

arnetheduck: what? `when not defined(orc/arc): proc baseType(...)` - there's no `{.error.}` here.. we should really change the `nimTypeNames` runtime assert to compile-time as well, they're the same

Menduist: It works if I switch to `proc baseType*[T: RootObj](obj: T): cstring {.deprecated.} =` (since the error will be done at instantiation), but not sure we want do to that

Menduist: Switched to generic, why not

ringabout: Is there something wrong with https://github.com/status-im/nim-stew/pull/148#discussion_r1017974909? It breaks at compile time and gives a nicer messages for ARC/ORC.

```
Error: baseType is not available in ARC/ORC; usage of 'baseType' is an {.error.} defined at
```

ringabout: You can also include `nimTypeNames` for defines and gives a proper error message.

Menduist: Sorry github (or my brain) derped and didn't show your comment. Better indeed, thanks

ringabout:
BTW, the previous CI consumes lots of resources, and now it is queued. How about only running the CI on the latest commit and cancel the previous ones. The Nim repo has been doing this without a hitch.

ref https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/72408109#72408109

arnetheduck:
> BTW, the previous CI consumes lots of resources,

we'll happily take patches for this, though in the case of stew the jobs are pretty fast so it hasn't been a prio fixing it - it's a good idea in general though, indeed

arnetheduck (approved):

Issue #151, opened 2022-11-15
after change lib/system/arithmetics.nim proc inc to generic, the proc inc in this module cause  ambiguous call


bung87:
see https://github.com/nim-lang/Nim/actions/runs/3463825054/jobs/5784591531

https://github.com/status-im/nim-stew/blob/1e86bd1ef38f78c601b07da7188e65785f2c0ed8/stew/interval_set.nim#L330

Menduist:
What's the purpose of your linked PR?

bung87:
the purpose is really make targeting compile target system int bit width, check `compiler/sigmatch.nim` changes.  before is match `tyInt` to `tyInt32`.

arnetheduck:
While the `inc` function in `interval_set` looks like it could use a round of improvement indeed (ie it has overly broad matching and dubious utility vs `+=` cc @mjfh ), it's generally a good idea to avoid generic functions in core stdlib functions - if anything, you can overload instead which doesn't trigger nearly as many type resolution issues (and sometimes serve as "type anchors")

bung87:
yeah, I agree, the orignal version already generic the poblem part is second argument `y = 1` which now matches to `tyInt`

mjfh:
the `inc()` will be replaced by `incPt()` which should do the job

Issue #153, opened 2022-11-17
sorted set tests broken when run individually


arnetheduck:
```
nim c -r test_sorted_set "Delete items"

[Suite] SortedSet: Sorted list based on red-black tree
    /home/arnetheduck/status/nimbus-eth2/vendor/nim-stew/tests/test_sorted_set.nim(133, 22): Check failed: data.isOk == canDeleteOk
    data.isOk was false
    canDeleteOk was true
/home/arnetheduck/status/nimbus-eth2/vendor/nim-unittest2/unittest2.nim(848) unittest2
/home/arnetheduck/status/nimbus-eth2/vendor/nim-stew/tests/test_sorted_set.nim(136) runTestX60gensym287
/home/arnetheduck/status/nimbus-eth2/vendor/nim-stew/stew/results.nim(378) get
/home/arnetheduck/status/nimbus-eth2/vendor/nim-stew/stew/results.nim(368) raiseResultDefect

    Unhandled exception: Trying to access value with err Result: rbEmptyTree [ResultDefect]
```

this is because a global variable is used here:

https://github.com/status-im/nim-stew/blob/master/tests/test_sorted_set.nim#L76

suite variables should use `setup:` to ensure that a fresh variable is used for every test

arnetheduck:
> this is exactly what is not used here, one test are dependent 

tests should generally be independent unless there's a good reason - ie if you want tests to share data, they should do so explicitly via an initialization that gets called on every test start, not accidentally by relying on the order in which the test runner sometimes runs them

mjfh:
Understood, nevertheless there is some merit in writing dependent tests reporting as a single section. What I can do here is skipping the last test if it is not feasible.

Glueing together all dependent tests into a single one need there own output in sort of a subsections IMHO. is there such a feature?

arnetheduck:
> e is some merit in writing dependent tests reporting as a single section.

the right technique for doing that is to move the initialization code to a separate function and call that function from both tests (or indeed put them under one `test`).

tests are not guaranteed to be run in order by the test running - for example, parallel test runners run them in random order - ditto tools that re-run failed tests automatically - introducing order-dependence in tests is undefined behaviour, essentially.

PR #154, opened 2022-11-17
Fix issues #151 and #153

mjfh:


tests/test_sorted_set.nim
```diff
@@ -117,25 +117,28 @@ suite "SortedSet: Sorted list based on red-black tree":
   test "Delete items":
     var seen: seq[int]
     let sub7 = keyList.len div 7
-    for n in toSeq(countup(0,sub7)).concat(toSeq(countup(3*sub7,4*sub7))):
-      let
-        key = keyList[n]
-        canDeleteOk = (key notin seen)
+    if sl.len == 0:
```
arnetheduck, line 120: this doesn't really fix the issue: it skips the test where in reality, when running the one test by itself, it should still perform the same test as before: the way to do this is to move the initialization code that sets up the test into a function and call it from both delete items and the other test

PR #157, opened 2022-11-18
byteutils: allow empty output buffers

arnetheduck:
This PR supercedes #144 to allow empty outputs in hex parsing

### DIFF
--- a/stew/byteutils.nim
+++ b/stew/byteutils.nim
@@ -40,77 +40,122 @@ template skip0xPrefix(hexStr: openArray[char]): int =
   if hexStr.len > 1 and hexStr[0] == '0' and hexStr[1] in {'x', 'X'}: 2
   else: 0
 
-func hexToByteArrayImpl(hexStr: openArray[char], output: var openArray[byte], fromIdx, toIdx: int): int
-                       {.raises: [ValueError, Defect].} =
-  ## Read a hex string and store it in a byte array `output`. No "endianness" reordering is done.
-  ## Allows specifying the byte range to process into the array
+func hexToByteArrayImpl(
+    hexStr: openArray[char], output: var openArray[byte], fromIdx, toIdx: int):
+    int {.raises: [ValueError, Defect].} =
   var sIdx = skip0xPrefix(hexStr)
+  # Fun with open intervals
+  doAssert fromIdx >= 0 and
+    toIdx <= output.high and
+    fromIdx <= (toIdx + 1)
 
-  doAssert(fromIdx >= 0 and toIdx >= fromIdx and fromIdx < output.len and toIdx < output.len)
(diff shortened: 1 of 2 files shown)


stew/byteutils.nim
```diff
@@ -40,77 +40,122 @@ template skip0xPrefix(hexStr: openArray[char]): int =
...
-                       {.raises: [ValueError, Defect].} =
-  ## Read a hex string and store it in a byte array `output`. No "endianness" reordering is done.
-  ## Allows specifying the byte range to process into the array
+func hexToByteArrayImpl(
+    hexStr: openArray[char], output: var openArray[byte], fromIdx, toIdx: int):
+    int {.raises: [ValueError, Defect].} =
   var sIdx = skip0xPrefix(hexStr)
+  # Fun with closed intervals
+  doAssert fromIdx >= 0 and
+    toIdx <= output.high and
```
zah, line 49: `output.high` involves arithmetic. Why did you decide to change the previous version based on `output.len`?

arnetheduck: because fromIdx and toIdx are closed intervals, and comparing closed intervals with half-open ones is a PITA. 

-1 is a valid `toIdx` - if there is static analysis done, it needs to take this into account.

zah: I meant that the `output.high` expression is semantically dubious when `output` is empty. It would be reasonable for a static analyzer to produce a warning on it. To be clear, I'm just suggesting that you keep the old `toIdx < output.len` which is exactly the same check (`high` gets translated to `output.len - 1`).

arnetheduck: hm, this is the idiomatic nim way after all: `toOpenArray(a.low, a.high)` is how you create one so using `high` just reflects the natural use of the "constructor" of an openarray. the fact that underneath there's a lot of +1 and -1 happening is a sad consequence, but it is what it is.

>  It would be reasonable for a static analyzer to produce a warning on it. 

the entire nim ecosystem is built up around the design choice that an empty interval is `[0, -1]` while `[0, 0]` is a 1-element construct - any static analyzer we use will have to understand this since every time you write `0..<len`, -1 is happens as well (it would even be correct to claim that it would be better "nim" to use `0..high` consistently). I'm with dijkstra and you on this topic, philosophically: it would indeed be reasonable for a static analyzer to point this out but we'll have to silence that warning because of the way the language was construed.

Also, I don't really mind changing this, but the mixing of closed and open ranges causes bugs - any sane programmers will naturally tend towards len and half-open, but here we have to make a choice of fighting the language or going with the flow, and the latter tends to win in the end - above all, we'd have to go with a slice type that works with start+len rather than start/end and that's a long journey - until we've done that, the least bug-prone way I know of to compare closed ranges is with `>=` and `<=` on both ends.

PR #162, opened 2022-11-28
ensure `isZeroMemory` used only on types without tuple/object padding

tersec:
Otherwise, `isZeroMemory` won't necessarily correspond to the Nim notion of being a default-initialized object of that type.


zah:
Using `supportsCopyMem` is probably the right way to solve this.

tersec:
> Using `typetraits.supportsCopyMem` is probably the right way to solve this.

Well, maybe `supportsEqualMem`: https://github.com/nim-lang/Nim/issues/19897

tersec:
That doesn't exist, though, so the intent is to at least flag obviously-incorrect usages.

PR #165, opened 2023-01-19
Remove lock file

arnetheduck:
The lock file breaks devel testing - CI updates are needed as well as a strategy for handling "nim important packages" builds (which build with a version of Nim different from that in the lock file)

### DIFF
--- a/nimble.lock
+++ b/nimble.lock
@@ -1,4 +1,0 @@
-{
-  "version": 1,
-  "packages": {}
-}


zah:
Can you elaborate on how the lockfile breaks Nim devel testing?

Even after Nim is locked in the lockfile, nimble commands such as `nimble build` have the option `--use-system-nim` which tells Nimble to use the first Nim executable found in the system PATH, so I would argue the Nim important packages use case is fully supported.

arnetheduck:
on our end, we've had CI build failures because of this for weeks without anyone taking action, and it's disrupting upstream's testing as well - the lock files have also been without maintainer for months - when there's maintenance in place and a strategy for handling the ensuing lock file churn, we can readd them but until then they disrupt day-to-day work (I'm happy to close this if a PR comes along that fixes all of the above, but until then..): 

* https://github.com/status-im/nim-stew/actions/runs/3956216159/jobs/6775208041
* https://github.com/status-im/nim-chronos/actions/runs/3951853680/jobs/6778673427
* https://github.com/status-im/nim-stew/commit/32d6f43fffa09b85e88c89f2b61cb96aea544737#commitcomment-97319913

arnetheduck:
more of the same: https://github.com/status-im/nim-chronos/issues/337 (and similar issues elsewhere)

zah:
OK, I agree that the lockfiles need to be maintained and it's reasonable to temporarily remove them until the team starts to use the intended tooling. I've tasked @yyoncho with delivering an exemplar Nimble integration in the CI of LibP2P repo which will hopefully resolve some of the remaining contention points.

There was another way to "fix" the issue here though. You could have tried the new Nimble and/or the nim-workspace workflow and produced an up-to-date lockfile.

arnetheduck:
> produced an up-to-date lockfile.

not quite, we need separate lock files for every nim version _and_ significant updates to the CI scripts _and_ a working solution for the upstream "important packages" for practically every project - this is not a 5-minute job but rather a complex task that requires dedicated time and effort.

PR #171, opened 2023-02-14
reduce compiler warnings

jangko:


stew/byteutils.nim
```diff
@@ -25,7 +25,7 @@ else:
 #####################################   Hex utilities   ################################################
 
 proc readHexChar*(c: char): byte
-                 {.raises: [ValueError, Defect], noSideEffect, inline.} =
```
arnetheduck, line 28: we cannot remove `Defect` here until we drop 1.2 support, which, for stew, will happen last among all our libraries.

the alternative is to use a macro akin to https://github.com/markspanbroek/upraises/blob/main/upraises.nim#L52 - it's an open question whether it's worth the complexity

jangko: I've reenable it using `pragma pragma` instead of using macro.

PR #176, opened 2023-04-13
results: Add `isOkOr`, `isErrOr`

arnetheduck:
These two helpers complete `valueOr` and `errorOr` to cover `void` cases where no value should be returned or `Result[void, E]` / `Result[T, void]` is being used - they can be used for a convient early-return style in side-effectful proc:s:

```nim
v.update().isOkOr:
  echo "update failed: ", error
```


kdeme:
Great, I've hit that use case several times already :)

PR #186, opened 2023-05-09
remove cast warnings

diegomrsantos:


stew/bitops2.nim
```diff
@@ -82,7 +82,7 @@ func log2truncNim(x: uint8|uint16|uint32): int =
   v = v or v shr 4
   v = v or v shr 8
   v = v or v shr 16
-  cast[int](lookup[uint32(v * 0x07C4ACDD'u32) shr 27])
```
arnetheduck, line 85: the cast here is intentional to avoid a runtime check of the returned value - so this should be `cast[int](uint(lookup...` - ditto other `int` casts below

Menduist: Maybe lookup could store ints directly? Would just cost a bit of memory

Menduist: Isn't the compiler smart enough to not put checks that are impossible? uint8->int is always safe and doesn't require checks

arnetheduck: ah good point, if this indeed works - it _should_ be able to do it, but that's one side of the coin only ;)

tersec: Well, if it doesn't, and there's a reproducible test case, that sounds like a good candidate for a Nim bug to file. In the meantime, the correctness per se of removing casts wouldn't be in question, just efficiency, in versions of Nim which don't receive said backports.

arnetheduck: > just efficiency

I agree with you on this one _except_ in modules whose raison d'etre is efficiency - like bitops.

Menduist: For https://github.com/status-im/nim-libp2p/pull/813 we added a [`safeConvert`](https://github.com/status-im/nim-libp2p/pull/813/files#diff-f53255fc8b92e8a57093f8bf53071913ca7d3deed5ef29f7f05bf89ecd3e02e0R69-R74) that checks that no defects (or junk values) are possible:
`safeConvert[int](xxx)`

It turned out quite useful, already found a few bugs, could be moved to stew

diegomrsantos: My motivation was this comment https://github.com/nim-lang/Nim/pull/20103#issuecomment-1225986280.

Is type cast for different number of bits well-defined in Nim?

diegomrsantos: How about `const lookup: array[32, int]`?

arnetheduck: > safeConvert[int](xxx)

https://github.com/status-im/nim-stew/pull/34 - this would / should be covered by the "lossless conversion" part of that framework - the idea generalises quite well.

I'm somewhat against putting "safe" in names frivolously without defining "safe" generally in the language - it gets overloaded for a bunch of confusing and sometimes contradictory purposes since "safe" on its own has no real meaning.

Since the language already warns against "unsafe" usage of conversions specifically, all we need to do is to turn the warning into an error once all known uses are fixed.

> Is type cast for different number of bits well-defined in Nim?

no - that's why there's a warning for it now - casting to a larger type is UB (in other words: if ever you manage to cast to a larger type without a warning being printed, it's a bug that should be reported upstream).

arnetheduck: > Since the language already warns against "unsafe" usage of conversions specifically, all we need to do is to turn the warning into an error once all known uses are fixed.

Oops, right, that was silly: it doesn't warn in all cases, specifically the ones we're discussing here :)

Ok, so I agree a construct would be useful that would allow the defect-free conversions.

arnetheduck (approved):

stew/bitops2.nim
Menduist: > this would / should be covered by the "lossless conversion" part of that framework

:+1:, would definitively be nice to have this framework set up, could also include the work that was done on enum https://github.com/status-im/nim-stew/pull/115

PR #187, opened 2023-05-09
intops: core integer primitives

arnetheduck:
This, together with bitops2 and endians2, forms the core primitive offering for working with integers as the computer sees them.

The focus of intops is to expose a number of common integer operations typically used to build more complex abstractions such as bigints, mp-ints etc while having access to the best performance the compiler and cpu can offer.

There is more to do here, but this provides an outline of what this module could look like.

Obviously, there are no exceptions or defects around - the point of these utilities is to stay as close as possible to bare metal. They could be used to implement such features however (similar to how `system/integerops` works).

### DIFF
--- a/stew/intops.nim
+++ b/stew/intops.nim
@@ -1,0 +1,122 @@
+## Core integer primitives suitable as building blocks for higher-level
+## functionality such as bigints, saturating integer types etc - where
+## applicable, these use compiler builtins - otherwise, they fall back on native
+## Nim code that may be less efficient.
+##
+## In using these functions, it is recommended that you always call the function
+## that returns the least information needed - for example, `mulOverflow` may
+## be implemented more efficiently than `mulWiden`, meaning that if overflow
+## detection is all that is needed, use the former.
+
+# Implementation notes:
+#
+# * `uintN` is assumed to be wrapping
+# * "*Overflow" perform wrapping arithmetic while returning a bool for overflow
+# * "*Widen" return full result in multiple words
+# * overloads with carry/borrow exposed for chaining limbs
+#
+# TODO
+# * use compiler intrinsics
+# * signed ops
+# * saturating ops
+# * more primitives commonly available on CPU:s / intrinsics (pow / divmod / etc)
+
+
+func addOverflow*(x, y: SomeUnsignedInt):
+    tuple[result: SomeUnsignedInt, overflow: bool] =
+  ## Add the two integers using wrapping arithmetic, returning the result and a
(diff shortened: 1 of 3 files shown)


stew/intops.nim
```diff
@@ -0,0 +1,133 @@
...
+# * more primitives commonly available on CPU:s / intrinsics (pow / divmod / etc)
+# * discovery mechanism to determine implementation quality
+#
+# References:
+# https://llvm.org/docs/LangRef.html#arithmetic-with-overflow-intrinsics
+# https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
+# https://doc.rust-lang.org/std/primitive.u32.html#implementations
+
+func addOverflow*(x, y: SomeUnsignedInt):
+    tuple[result: SomeUnsignedInt, overflow: bool] =
```
zah, line 37: Does this really simplify the implementation of BigInt libraries? If I try to imagine the loop that will be used there, it seems to me that it will be more complicated and less performant when based on this helper function.

In particular, the reliance on a tuple that gets translated to a C struct is what makes me nervous. If the carry is communicated with an output parameter, the compiler is a bit more free to perform register allocations in more optimal ways.

Also, ultimately, the carry should probably be obtained from the CPU itself, but I guess your plan is to replace the bodies of these funds in the future?

arnetheduck: >  If the carry is communicated with an output parameter, the compiler is a bit more free to perform register allocations in more optimal ways.

Typically modern compilers are able to deal with this - also, the fact that it's a return value and not a pointer gives the compiler some freedoms it otherwise doesn't have - in llvm for example, this is typically handled by the SROA pass that decomposes structs into individual elements then assigns registers based on the lifetimes of the fields themselves (which in this case are trivial).

see https://gcc.godbolt.org/z/Ex8P76fWr for an example of how it works with a struct ret type.

These implementations are meant for the VM mainly - the actual (future) implementations would use compiler builtins which unfortunately differ in their actual API, but yes, the ideal is that the compiler maps a function like this to its ADC instruction that does a 3-operand addition returning the carry in a flag.

For bigints, the 3-parameter carry form in particular is interesting - for saturating arithmetic, the 2-paremeter version without carry is more appropriate.

arnetheduck: Fixed a bug and improved example: the code now shows that using the builtin actually has no advantage over the no-builtin code in this particular case - both end up using the right `ADC` instruction for a 192-bit integer for example:

https://gcc.godbolt.org/z/bvW6aTr5a

PR #198, opened 2023-06-12
fix `ProveField` warnings in `results`

etan-status:
Nim emits `ProveField` warnings with `if` on case object discriminator. Replace with `case` instead to avoid those warnings. Note we currently have `ProveField` disabled but it keeps showing up sometimes when compiling with `nim c` instead of `make`.

### DIFF
--- a/stew/results.nim
+++ b/stew/results.nim
@@ -377,13 +377,16 @@ func raiseResultDefect(m: string, v: auto) {.noreturn, noinline.} =
 func raiseResultDefect(m: string) {.noreturn, noinline.} =
   raise (ref ResultDefect)(msg: m)
 
-template assertOk(self: Result) =
+template withAssertOk(self: Result, body: untyped): untyped =
   # Careful - `self` evaluated multiple times, which is fine in all current uses
-  if not self.oResultPrivate:
+  case self.oResultPrivate
+  of false:
     when self.E isnot void:
       raiseResultDefect("Trying to access value with err Result", self.eResultPrivate)
     else:
       raiseResultDefect("Trying to access value with err Result")
+  of true:
+    body
 
 template ok*[T: not void, E](R: type Result[T, E], x: untyped): R =
   ## Initialize a result with a success and value
@@ -456,13 +459,14 @@ func map*[T0: not void, E; T1: not void](
   ## let r = Result[int, cstring).ok(42)
   ## assert r.map(proc (v: int): int = $v).value() == "42"
   ## ```
-  if self.oResultPrivate:
+  case self.oResultPrivate
+  of true:
     when T1 is void:
       f(self.vResultPrivate)
       result.ok()
     else:
       result.ok(f(self.vResultPrivate))
-  else:
+  of false:
     when E is void:
       result.err()
     else:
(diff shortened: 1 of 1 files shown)


etan-status:
Note, this doesn't get rid of _all_ `ProveFields` warnings, as some need to suppress it at call side (e.g., `var`).

However, it still, overall, reduces the amount of warnings quite a lot, so is possibly a net gain overall.

etan-status:
Nim devel 🤔 (seems to be alright on `master`)

```sh
/home/runner/.nimble/pkgs2/unittest2-0.0.7-b6d4a5cbe28b43c166d6442ba6804aafd4abe368/unittest2.nim(900) test_interval_set
/home/runner/work/nim-stew/nim-stew/tests/test_interval_set.nim(129) runTest`gensym123
/home/runner/work/nim-stew/nim-stew/stew/interval_set.nim(1164) verify
/home/runner/work/nim-stew/nim-stew/stew/interval_set.nim(345) new
/home/runner/work/nim-stew/nim-stew/stew/interval_set.nim(295) right
/home/runner/work/nim-stew/nim-stew/stew/interval_set.nim(289) blk
/home/runner/work/nim-stew/nim-stew/nim/lib/system/arc.nim(43) nimIncRef
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

etan-status:
@araq re Nim `devel` issue. Something in this PR seems to mess with orc.

arnetheduck (approved):
lgtm but it feels like those devel problems should be investigated in depth before merging

tersec:
Nim `devel` issues should not block development progress. I stand by what I wrote in https://github.com/status-im/nim-ssz-serialization/pull/47#issuecomment-1584890664
> But it remains that the primary purpose of the `nim-ssz-serialization` CI is to make `nim-ssz-development` in the context of supported Nim versions (of which constantly-changing `devel` is not one), so if `devel` in the future causes too-regular CI regressions, it might prove warranted to reconsider this situation.

`s/nim-ssz-serialization/nim-stew/g`, but otherwise identical.

etan-status:
@ringabout Reran on Nim `fda8b6f193e2e229488f76f18089f01eb08272fb` which includes https://github.com/nim-lang/Nim/pull/22088

It's still broken here :) So, likely a separate bug.

ringabout:
Seems to be a 2.0 to devel regression.

arnetheduck:
> Nim devel issues should not block development progress. 

they shouldn't, but we can also use the opportunity to rid upstream of debilitating bugs that might come to bite us later - ie this is not an urgent PR in the sense that there's no runtime bug being fixed so we can merge it when convenient

tersec:
> > Nim devel issues should not block development progress.
> 
> they shouldn't, but we can also use the opportunity to rid upstream of debilitating bugs that might come to bite us later - ie this is not an urgent PR in the sense that there's no runtime bug being fixed so we can merge it when convenient

These can happen simultaneously: if/as useful, create a branch, destined never to be merged, which can function as a repro for the bug in question. Then fix `master`. Then both activities can happen in parallel: upstream Nim can fix as best suits them, and meanwhile, `nim-stew` development isn't hindered.

ringabout:
I'm fixing it, seems to be a piece of cake.

etan-status:
Looks good now on devel

PR #203, opened 2023-07-04
fix results compiletime regression

### LINKED ISSUE #202: withAssertOk introduce compile time regression of Results[T, E]
PR #198 change `assertOk` to `withAssertOk`, and it make compiletime regression of Results.

```Nim
import 
  ../stew/results

func bug(): Result[uint16, string] =
  ok(1234)
  
const
  x = bug()
  y = x.value()
  
echo y
```

Current output: ```0```
Expected output: ```1234```

If we change const to let, it will works as expected. 
It maybe a Nim bug, but I think we should make it works again while wait for nim compiler get fixed.

See also status-im/nimbus-eth1#1617

ping @etan-status

jangko:
fix #202

### DIFF
--- a/stew/results.nim
+++ b/stew/results.nim
@@ -801,7 +801,10 @@ func value*[T, E](self: Result[T, E]): T {.inline.} =
   ## See also: Option.get
   withAssertOk(self):
     when T isnot void:
-      self.vResultPrivate
+      # TODO: remove result usage.
+      # A workaround for nim VM bug:
+      # https://github.com/nim-lang/Nim/issues/22216
+      result = self.vResultPrivate
 
 func value*[T: not void, E](self: var Result[T, E]): var T {.inline.} =
   ## Fetch value of result if set, or raise Defect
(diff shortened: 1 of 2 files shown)


jangko:
hmm, turn out the other `func value` also regress for nim 1.2 and 1.4 but not tested before.

tersec:
I'm all for green CI, but, in this case, is this just creating silent brokenness in pre-Nim-1.6?

My suggested response to https://github.com/status-im/nim-stew/pull/203/commits/1b2c0be6a00a9f05c8c6c14bc2e2f8ba77cb4b26 is more along the lines of https://github.com/status-im/nim-stew/pull/204 -- if it's known broken without feasibility of fixing it, then it's not really a supported target.

And, while it might be possible to fix this so that it still works on Nim 1.2/1.4, just, why?

arnetheduck:
>  is this just creating silent brokenness in pre-Nim-1.6?

this is an example of partial support - ie most of it works but compile-time has issues - worth keeping in mind here is that when we talk about supporting 1.6, _most_ of the 1.6 releases actually don't work (only the newer ones).

etan-status (commented):
Thanks for catching this one!

PR #207, opened 2023-07-16
avoid zeroMem

arnetheduck:
on cached data, it doubles read time (!)


arnetheduck:
motivating example of reading a 100mb state:

image

tersec (approved):

PR #222, opened 2024-06-18
test: use Nimble to manage Nim

romanzac:
Test PR to understand wether we can reliably use Nimble to manage Nim.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -147,8 +147,8 @@ jobs:
           echo $PATH
           ls -la /home/runner/.nimble/pkgs2/nim-2.0.6-c43fce95a734c8ad12475e1ff10f522b408fbcfa/bin/nim
           /home/runner/.nimble/pkgs2/nim-2.0.6-c43fce95a734c8ad12475e1ff10f522b408fbcfa/bin/nim --version
-#          nim --version
           nimble --version
+#          nim --version
 #          nimble install -y --depsOnly
 #          rm -f nimble.lock
 #          env NIMLANG=c nimble test
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -139,8 +139,8 @@ jobs:
           curl -O -L -s -S https://github.com/nim-lang/nimble/releases/download/latest/nimble-linux_x64.tar.gz
           tar xvfz nimble-linux_x64.tar.gz
           ./nimble install nim
+          echo $(./nimble path nim)/bin
           echo $(./nimble path nim)/bin >> $GITHUB_PATH
-          echo $GITHUB_PATH
 
       - name: Run tests
         run: |
@@ -148,7 +148,7 @@ jobs:
             # https://github.com/status-im/nimbus-eth2/issues/3121
             export NIMFLAGS="-d:nimRawSetjmp"
           fi
-          echo $GITHUB_PATH
+          echo $PATH
           nim --version
           nimble --version
           nimble install -y --depsOnly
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,152 @@
...
+          esac
+          [[ -z "$ncpu" || $ncpu -le 0 ]] && ncpu=1
+          echo "ncpu=$ncpu" >> $GITHUB_ENV
+          echo "MAKE_CMD=${MAKE_CMD}" >> $GITHUB_ENV
+
+      - name: Download Nimble and install Nim
+        run: |
+          curl -O -L -s -S https://github.com/nim-lang/nimble/releases/download/latest/nimble-linux_x64.tar.gz
+          tar xvfz nimble-linux_x64.tar.gz
+          ./nimble install nim
```
arnetheduck, line 141: broadly, we should be able to use `nimble setup` here, which _should_ install a package-local version of nim that `nimble test` later uses

romanzac: Nimble setup worked out https://github.com/status-im/nim-stew/actions/runs/9663859361/job/26657108122

romanzac: nimble install nim@1.6.20 also worked out when full version tag was specified.
https://github.com/status-im/nim-stew/actions/runs/9667227572/job/26668548913

romanzac: nimble install nim@2.1.1 ended well, but similar situation with older Nim. This time instead of Nim 2.1.1 older Nim 2.0.6 was effectively used to run the tests.
https://github.com/status-im/nim-stew/actions/runs/9667421954/job/26669209430

romanzac: What would we like to see here ? I would personally see the easiest to have Nimble which could install older Nim, and fix versions of older Nim which could follow reasonably young Nimble through back porting. 

Example:
Nim 1.6 is the last major.minor for v1 and its fix versions 1.6.20 and so on would allow being managed by Nimble which could manage Nim v2 versions.

This would relieve devs who need to stick to Nim 1.6 for very long time. Maybe could add 2+ years of lifetime?

If this is too complex or slow on innovation, we might follow the rustc and cargo approach. Moving faster through versions and having same versioning for compiler and package manager. Rust has also editions to create "promise" across many versions. Not sure if any promise is kept for Nim major versions ?

romanzac: I've found a discussion on the related topic in "very old high priority" feature request 654 at issues at nimble at minlang. I feel like we would need to start collecting numbers and data points to give sea of opinions clarity they deserve. :)

romanzac: I've compiled report about what I am doing. To tidy up and have something presentable for community. https://www.notion.so/Nim-Tooling-Test-Reports-7b25687860ae41afaa2d4f22af8d07f1?pvs=4 

What should we do next ? @arnetheduck

romanzac: Add Tests05 and Tests06 to compare download and installation/setup time to the report.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          if [[ '${{ matrix.target.cpu }}' == 'amd64' ]]; then
+            PLATFORM=x64
+          else
+            PLATFORM=x86
+          fi
+          echo "PLATFORM=$PLATFORM" >> $GITHUB_ENV
+
+          ncpu=
+          MAKE_CMD="make"
```
arnetheduck, line 120: these make sections are not needed

romanzac: Aren't they needed for Windows ? 

Right now this PR is only to test and collect information. In the [test report](https://www.notion.so/Nim-Tooling-Test-Reports-7b25687860ae41afaa2d4f22af8d07f1), there is only one passing combination - Test01. When Nimble installs Nim 2.0 and Nimble 0.14.2.

arnetheduck:
obsoleted by the move to the common workflow (which indeed could learn to use nimble to install nim - which is probably feasible now)

PR #243, opened 2025-04-23
feat: add base36

gmelodie:
Added base36 encoding and decoding

### DIFF
--- a/stew/base36.nim
+++ b/stew/base36.nim
@@ -1,0 +1,223 @@
+## Copyright (c) 2025 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+## This module implements Base36 encoding and decoding procedures.
+## This module supports two Base36 alphabets: lowercase `Base36Lc` and uppercase `Base36Uc`.
+
+## This module follows Base36 as defined in
+# https://github.com/multiformats/multibase/blob/f378d3427fe125057facdbac936c4215cc777920/rfcs/Base36.md
+
+{.push raises: [].}
+
+type
+  Base36Status* {.pure.} = enum
+    Error,
+    Success,
+    Incorrect,
+    Overrun
+
+  Base36Alphabet* = object
+    decode*: array[128, int8]
+    encode*: array[36, uint8]
+
+  Base36* = object
+  Base36Uc* = object
+  Base36Lc* = object
+
+  Base36Types* = Base36 | Base36Uc | Base36Lc
+
+  Base36Error* = object of CatchableError
+    ## Base36 specific exception type
+
+func newAlphabet36(s: static[string]): Base36Alphabet =
+  doAssert(len(s) == 36)
+  var alphabet: Base36Alphabet
+  for i in 0..<len(s):
+    alphabet.encode[i] = cast[uint8](s[i])
+  for i in 0..<len(alphabet.decode):
+    alphabet.decode[i] = -1
+  for i in 0..<len(alphabet.encode):
+    alphabet.decode[int(alphabet.encode[i])] = int8(i)
+  return alphabet
+
+const
+  B36UcAlphabet* = newAlphabet36("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+  B36LcAlphabet* = newAlphabet36("0123456789abcdefghijklmnopqrstuvwxyz")
+
+func encodedLength*(btype: typedesc[Base36Types], length: int): int =
+  ## Return estimated length of Base36 encoded value for plain length
+  ## ``length``.
+  return (length * 277) div 179 + 1
+
+func decodedLength*(btype: typedesc[Base36Types], length: int): int =
+  ## Return estimated length of decoded value of Base36 encoded value of length
+  ## ``length``.
+  return (length * 179) div 277
+
+func encode*(btype: typedesc[Base36Types], inbytes: openArray[byte],
+             outstr: var openArray[char], outlen: var int): Base36Status =
+  when (btype is Base36) or (btype is Base36Lc):
+    const alphabet = B36LcAlphabet
+  elif (btype is Base36Uc):
+    const alphabet = B36UcAlphabet
+
+  let binsz = inbytes.len
+  var zcount = 0
+  while zcount < binsz and inbytes[zcount] == 0x00'u8:
+    inc(zcount)
+
+  let size = btype.encodedLength(binsz - zcount)
+  var buffer = newSeq[uint8](size)
+
+  var hi = size - 1
+  for i in zcount..<binsz:
+    var carry = uint32(inbytes[i])
+    var j = size - 1
+    while (j > hi) or (carry != 0'u32):
+      carry = carry + uint32(256'u32 * buffer[j])
+      buffer[j] = cast[byte](carry mod 36)
+      carry = carry div 36
+      dec(j)
+    hi = j
+
+  # Strip leading zeros in encoded buffer
+  var j = 0
+  while j < size and buffer[j] == 0x00'u8:
+    inc(j)
+
+  outlen = 1 + zcount + (size - j) # 1 for 'k'
+  if outstr.len < outlen:
+    return Base36Status.Overrun
+
+  outstr[0] = 'k'
+  for k in 0..<zcount:
+    outstr[1 + k] = cast[char](alphabet.encode[0])
+
+  var i = 1 + zcount
+  while j < size:
+    outstr[i] = cast[char](alphabet.encode[buffer[j]])
+    inc(j)
+    inc(i)
+
+  return Base36Status.Success
+
+
+func encode*(btype: typedesc[Base36Types],
+             inbytes: openArray[byte]): string {.inline.} =
+  ## Encode array of bytes ``inbytes`` using Base36 encoding and return
+  ## encoded string.
+  var size = btype.encodedLength(inbytes.len) + 1
+  var encoded = newString(size)
+  if btype.encode(inbytes, encoded.toOpenArray(0, size - 1),
+                  size) == Base36Status.Success:
+    encoded.setLen(size)
+  else:
+    encoded = ""
+  return encoded
+
+func decode*[T: byte|char](btype: typedesc[Base36Types], instr: openArray[T],
+             outbytes: var openArray[byte], outlen: var int): Base36Status =
+  when (btype is Base36) or (btype is Base36Lc):
+    const alphabet = B36LcAlphabet
+  elif (btype is Base36Uc):
+    const alphabet = B36UcAlphabet
+
+  if instr.len == 0 or instr[0] != 'k':
+    outlen = 0
+    return Base36Status.Incorrect
+
+  let payload = instr[1..^1]
+  let binsz = payload.len + 4
+  if outbytes.len < binsz:
+    outlen = binsz
+    return Base36Status.Overrun
+
+  var bytesleft = binsz mod 4
+  var zeromask: uint32
+  if bytesleft != 0:
+    zeromask = cast[uint32](0xFFFF_FFFF'u32 shl (bytesleft * 8))
+
+  let size = (binsz + 3) div 4
+  var buffer = newSeq[uint32](size)
+
+  var zcount = 0
+  # Handle leading zeros in the input string
+  while zcount < payload.len and payload[zcount] == cast[char](alphabet.encode[0]):
+    inc(zcount)
+
+  for i in zcount..<payload.len:
+    if (cast[byte](payload[i]) and 0x80'u8) != 0:
+      outlen = 0
+      return Base36Status.Incorrect
+    let ch = alphabet.decode[int8(payload[i])]
+    if ch < 0:
+      outlen = 0
+      return Base36Status.Incorrect
+    var c = uint32(ch)
+    for j in countdown(size - 1, 0):
+      let t = uint64(buffer[j]) * 36 + c
+      c = cast[uint32]((t and 0x3F_0000_0000'u64) shr 32)
+      buffer[j] = cast[uint32](t and 0xFFFF_FFFF'u32)
+    if c != 0:
+      outlen = 0
+      return Base36Status.Incorrect
+    if (buffer[0] and zeromask) != 0:
+      outlen = 0
+      return Base36Status.Incorrect
+
+  var boffset = 0
+  var joffset = 0
+  if bytesleft == 3:
+    outbytes[boffset] = cast[uint8]((buffer[0] and 0xFF_0000'u32) shr 16)
+    inc(boffset)
+    bytesleft = 2
+  if bytesleft == 2:
+    outbytes[boffset] = cast[uint8]((buffer[0] and 0xFF00'u32) shr 8)
+    inc(boffset)
+    bytesleft = 1
+  if bytesleft == 1:
+    outbytes[boffset] = cast[uint8]((buffer[0] and 0xFF'u32))
+    inc(boffset)
+    joffset = 1
+
+  while joffset < size:
+    outbytes[boffset + 0] = cast[byte]((buffer[joffset] shr 0x18) and 0xFF)
+    outbytes[boffset + 1] = cast[byte]((buffer[joffset] shr 0x10) and 0xFF)
+    outbytes[boffset + 2] = cast[byte]((buffer[joffset] shr 0x8) and 0xFF)
+    outbytes[boffset + 3] = cast[byte](buffer[joffset] and 0xFF)
+    boffset += 4
+    inc(joffset)
+
+  outlen = binsz
+  var m = 0
+  while m < binsz:
+    if outbytes[m] != 0x00:
+      if zcount > m:
+        return Base36Status.Overrun
+      break
+    inc(m)
+    dec(outlen)
+
+  if m < binsz:
+    moveMem(addr outbytes[zcount], addr outbytes[binsz - outlen], outlen)
+  outlen += zcount
+
+  return Base36Status.Success
+
+
+func decode*(btype: typedesc[Base36Types], instr: string): seq[byte] {.raises: Base36Error.} =
+  ## Decode Base36 string ``instr`` and return sequence of bytes as result.
+  var decoded: seq[byte]
+  if instr.len > 0:
+    var size = instr.len + 4
+    decoded = newSeq[byte](size)
+    if btype.decode(instr, decoded, size) == Base36Status.Success:
+      decoded.setLen(size)
+    else:
+      raise newException(Base36Error, "Incorrect base36 string")
+  return decoded
--- a/tests/test_base36.nim
+++ b/tests/test_base36.nim
@@ -1,0 +1,97 @@
+import unittest2
+import ../stew/base36
+import  ../stew/byteutils
+
+func fromHex(a: string): seq[byte] =
+  doAssert(len(a) %% 2 == 0)
+  var hex: seq[byte]
+  if len(a) == 0:
+    hex = newSeq[byte]()
+  else:
+    hex = newSeq[byte](len(a) div 2)
+    hex = hexToSeqByte(a)
+  return hex
+
+const TestVectors = [
+  ["", "k"],
+  ["0001", "k01"],
+  ["0000ff", "k0073"],
+  ["61", "k2p"],
+  ["626262", "k3u736"],
+  ["636363", "k3vlur"],
+  ["73696d706c792061206c6f6e6720737472696e67", "kdhbvo69kp5joh9co6f1lm8zq4bpp5rb"],
+  ["aaeb15231dfceb60925886b67d065299925915aeb172c06647", "k7s4kro2wr10fsld6o2284s1a5w45h4f0yk63ryf"],
+  ["516b6fcd0f", "k4gnba1hr"],
+  ["bf4f89001e670274dd", "kkos4yabkyy7fml"],
+  ["572e4794", "ko6tok4"],
+  ["ecac89cad93923c02321", "k520229tco195fokh"],
+  ["10c8511e", "k4nmvq6"],
+  ["672a39cf0b8be9483302b26413b1784c159b357fafba", "kgxduar7o4l1hyevbs8p62izfwct9q6n3ju"],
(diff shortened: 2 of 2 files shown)


stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+
+  Base36Types* = Base36 | Base36Uc | Base36Lc
+
+  Base36Error* = object of CatchableError
+    ## Base36 specific exception type
+
+proc newAlphabet36*(s: string): Base36Alphabet =
+  doAssert(len(s) == 36)
+  for i in 0..<len(s):
+    result.encode[i] = cast[uint8](s[i])
```
tersec, line 35: https://status-im.github.io/nim-style-guide/language.result.html
> Avoid using `result` for returning values.
>
> Use expression-based return or explicit `return` keyword with a value

tests/test_base36.nim
```diff
@@ -0,0 +1,121 @@
...
+    check:
+      Base36.decode("=", decres, decsize) == Base36Status.Incorrect
+      decsize == 0
+      Base36.decode("@", decres, decsize) == Base36Status.Incorrect
+      decsize == 0
+      Base36.decode("g&", decres, decsize) == Base36Status.Incorrect
+      decsize == 0
+      Base36.decode("10%", decres, decsize) == Base36Status.Incorrect
+      decsize == 0
+
```
tersec, line 121: extra empty line?

tests/test_base36.nim
```diff
@@ -0,0 +1,121 @@
+import unittest2
+import ../stew/base36
+
+when defined(nimHasUsed): {.used.}
```
tersec, line 4: all versions of Nim `nim-stew` supports have `used`

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
+## Copyright (c) 2018 Status Research & Development GmbH
```
tersec, line 1: 2018?

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+  for i in 0..<len(result.decode):
+    result.decode[i] = -1
+  for i in 0..<len(result.encode):
+    result.decode[int(result.encode[i])] = int8(i)
+
+const
+  B36UcAlphabet* = newAlphabet36("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+  B36LcAlphabet* = newAlphabet36("0123456789abcdefghijklmnopqrstuvwxyz")
+
+proc encodedLength*(btype: typedesc[Base36Types], length: int): int =
```
tersec, line 45: https://status-im.github.io/nim-style-guide/language.proc.html#functions-and-procedures-languageproc
> Prefer `func` - use `proc` when side effects cannot conveniently be avoided.

Likewise with the rest of the `proc`s here. If they can't all be `func`, something is strange.

tests/test_base36.nim
```diff
@@ -0,0 +1,121 @@
...
+  ["bf4f89001e670274dd", "kos4yabkyy7fml"],
+  ["572e4794", "o6tok4"],
+  ["ecac89cad93923c02321", "520229tco195fokh"],
+  ["10c8511e", "4nmvq6"],
+  ["672a39cf0b8be9483302b26413b1784c159b357fafba", "gxduar7o4l1hyevbs8p62izfwct9q6n3ju"],
+  ["0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4f505152535455565758595a5b5c5d5e5f606162636465666768696a6b6c6d6e6f707172737475767778797a7b7c7d7e7f808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9fa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebfc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedfe0e1e2e3e4e5e6e7e8e9eaebecedeeeff0f1f2f3f4f5f6f7f8f9fafbfcfdfeff", "168swoi6iuzj4fbwknlnh695zl88v65qcfgnwrwepqcxb9dysmluowqahvt3r9gsc1v47ssxdivjda3nttl6r044pzz7zwhtgu2mkow5ts28x2mbwenh3wfz4s1sarspfhlrakvqrgpmzb66sgtz2lzbotl7r28wcq8925c747b44l60vrk3scrin4zvnwn7pdsukgo6lgjhu1nuwj7yt1h9ujpe3os17onsk7sp4ysmytu568do2tqetwnrmbxb2dtd8kqorcoakaizlm9svr8axe1acxfursz11nubrhighfd64yhmp99ucvzr944n8co01o4x64cmbd8be0hqbm2zy5uwe4uplc4sa50xajel4bkkxb1kh21pisna37eqwpbpq11ypr"]
+]
+
+suite "BASE36 encoding test suite":
```
tersec, line 56: BASE is not an acronym in this context

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+
+  if m < binsz:
+    moveMem(addr outbytes[zcount], addr outbytes[binsz - outlen], outlen)
+  outlen += zcount
+
+  result = Base36Status.Success
+
+
+proc decode*(btype: typedesc[Base36Types], instr: string): seq[byte] =
+  ## Decode BASE36 string ``instr`` and return sequence of bytes as result.
```
tersec, line 231: not an acronym

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+  doAssert(len(s) == 36)
+  for i in 0..<len(s):
+    result.encode[i] = cast[uint8](s[i])
+  for i in 0..<len(result.decode):
+    result.decode[i] = -1
+  for i in 0..<len(result.encode):
+    result.decode[int(result.encode[i])] = int8(i)
+
+const
+  B36UcAlphabet* = newAlphabet36("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
tersec, line 42: https://status-im.github.io/nim-style-guide/language.proc.html#functions-and-procedures-languageproc
> Avoid public functions and variables (`*`) that don't make up an intended part of public API.

tersec: Likewise for a lot of the functions and variables here, which are literally all public right now. Which ones are the actual intended APIs?

tests/test_base36.nim
```diff
@@ -0,0 +1,121 @@
+import unittest2
+import ../stew/base36
+
+when defined(nimHasUsed): {.used.}
+
+func hexToBytes*(a: string, result: var openArray[byte]) =
```
tersec, line 6: This usage of `result` is slightly confusing due to Nim otherwise treating `result` specially, probably better to use another variable name.

tests/test_base36.nim
```diff
@@ -0,0 +1,121 @@
...
+        r = r or (10 + ord(c) - ord('A'))
+      of '0'..'9':
+        r = r or (ord(c) - ord('0'))
+      else:
+        doAssert(false)
+      inc(i)
+    result[k] = r.byte
+
+func fromHex*(a: string): seq[byte] =
+  doAssert(len(a) %% 2 == 0)
```
tersec, line 33: `%%` is specifically unsigned modulo, is there a reason to use that instead of `mod` which isn't quasi-type-coercive? `len` returns a signed `int`, not an unsigned `int`.

stew/base36.nim
tersec: https://github.com/status-im/nim-stew/compare/fb5a09dc994acd7865451903d2852852bd48493d..3a7e91351f18cde1279dc4a48e73d8fe5d6f95fb

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+
+  Base36Error* = object of CatchableError
+    ## Base36 specific exception type
+
+proc newAlphabet36*(s: string): Base36Alphabet =
+  doAssert(len(s) == 36)
+  for i in 0..<len(s):
+    result.encode[i] = cast[uint8](s[i])
+  for i in 0..<len(result.decode):
+    result.decode[i] = -1
```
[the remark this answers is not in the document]
tersec: https://github.com/status-im/nim-stew/compare/fb5a09dc994acd7865451903d2852852bd48493d..3a7e91351f18cde1279dc4a48e73d8fe5d6f95fb

tests/test_base36.nim
tersec: https://github.com/status-im/nim-stew/compare/3a7e91351f18cde1279dc4a48e73d8fe5d6f95fb..89498a6349addf3205cfefff8642f3d6f0f748c0

tests/test_base36.nim
tersec: https://github.com/status-im/nim-stew/compare/3a7e91351f18cde1279dc4a48e73d8fe5d6f95fb..89498a6349addf3205cfefff8642f3d6f0f748c0

stew/base36.nim
tersec: https://github.com/status-im/nim-stew/compare/3a7e91351f18cde1279dc4a48e73d8fe5d6f95fb..89498a6349addf3205cfefff8642f3d6f0f748c0

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+    alphabet.decode[i] = -1
+  for i in 0..<len(alphabet.encode):
+    alphabet.decode[int(alphabet.encode[i])] = int8(i)
+  return alphabet
+
+const
+  B36UcAlphabet* = newAlphabet36("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+  B36LcAlphabet* = newAlphabet36("0123456789abcdefghijklmnopqrstuvwxyz")
+
+func encodedLength(btype: typedesc[Base36Types], length: int): int =
```
tersec, line 47: Neither `encodedLength` nor `decodedLength` appear to be used?

tests/test_base36.nim
```diff
@@ -0,0 +1,120 @@
...
+      of 'A'..'F':
+        r = r or (10 + ord(c) - ord('A'))
+      of '0'..'9':
+        r = r or (ord(c) - ord('0'))
+      else:
+        doAssert(false)
+      inc(i)
+    output[k] = r.byte
+
+func fromHex(a: string): seq[byte] =
```
tersec, line 30: maybe reimplements https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/byteutils.nim#L188-L199
and could just call it instead?

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
+## Copyright (c) 2025 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+## This module implements Base36 encoding and decoding procedures.
+## This module supports two variants of Base36 encoding (Bitcoin and Flickr).
```
tersec, line 10: I've been trying to find references for this point without success so far. Both seem to use Base58, not Base36?
- https://metacpan.org/pod/Encode::Base58 specifically describes itself as being "for Flickr short URI" purposes; and
- https://en.bitcoin.it/wiki/Base58Check_encoding

It would be useful to cite, via URLs, in the comment block, where this statement is coming from.

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+
+  while zcount < binsz and inbytes[zcount] == 0x00'u8:
+    inc(zcount)
+
+  let size = ((binsz - zcount) * 277) div 179 + 1
+  var buffer = newSeq[uint8](size)
+
+  var hi = size - 1
+  var i = zcount
+  var j = size - 1
```
tersec, line 82: What is this `j` doing? Either the loop body runs at least once or it doesn't.

If the `while i < binsz` loop body runs at least once, then unconditionally,
```nim
    j = size - 1
```
If the `while i < binsz` loop body is not run at least once, then regardless,
```nim
  j = 0
```

And regardless of what `j` does in the loop, if the loop body runs, it (see second case above) is immediately replaced by `j = 0`.

So I don't understand:
- why initialize var j = size - 1` at all, it seems pointless;
- why treat the `while i < binsz` loop `j` as the same `j` as the post`-while i < binsz` loop `j`? One has scope only within the loop and it would considerably clarify things to have that variable lifetime tied to this lexical `while i < binsz` scope, that it doesn't affect anything later in the `encode` function.

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+  while i < binsz:
```
tersec, line 83: isn't this just a `for i in zcount ..< binsz` loop?

It's not clear to me why desugar it to that extent, even if that's sort of what it compiles into.

It just unconditionally `inc(i)`'s exactly once each loop iteration, there's no tricky control flow where a counting `for` loop doesn't straightforwardly work because only sometimes increments its loop counter.

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+    hi = j
+    inc(i)
+
+  j = 0
+  while (j < size) and (buffer[j] == 0x00'u8):
+    inc(j)
+
+  let needed = zcount + size - j
+  outlen = needed
+  if len(outstr) < needed:
```
tersec, line 100: Why use the `needed` variable at all rather than directly assigning to `outlen`? I don't see what this adds over the simpler:
```nim
  outlen = zcount + size - j
  if len(outstr) < outlen:
```

stew/base36.nim
```diff
@@ -0,0 +1,238 @@
...
+    return Base36Status.Overrun
+  else:
+    for k in 0..<zcount:
+      outstr[k] = cast[char](alphabet.encode[0])
+    i = zcount
```
tersec, line 105: This would have to be adjusted if the first disguised `for` loop is transformed into an actual `for` loop because that loop variable wouldn't be in scope anymore, but that seems fine, since it just gets unconditionally replaced by `zcount` here anyway. It's effectively treated as a new variable.

stew/base36.nim
tersec: Oh. It's because you just copy/pasted this entire module from https://github.com/status-im/nim-stew/blob/master/stew/base58.nim

tersec:
Would be good to link to https://github.com/multiformats/multibase/blob/f378d3427fe125057facdbac936c4215cc777920/rfcs/Base36.md somewhere.

tersec:
https://status-im.github.io/nim-style-guide/errors.exceptions.html#exceptions-errorsexceptions
> Annotate each module at top-level (before imports):
>
>  `{.push raises: [].}`
>
> Use explicit `{.raises.}` annotation for each public (`*`) function.

Examples:
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/ctops.nim#L22
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/leb128.nim#L33
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/ptrops.nim#L20
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/assign2.nim#L5
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/byteutils.nim#L19-L20
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/objects.nim#L10
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/shims/sets.nim#L10
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_reset.nim#L16
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_desc.nim#L70
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/io2.nim#L13
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/arrayops.nim#L7
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/keyed_queue.nim#L25
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_delete.nim#L11
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/shims/macros.nim#L20
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/bitseqs.nim#L1
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_verify.nim#L11
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_insert.nim#L11
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_flush.nim#L17
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/endians2.nim#L21
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/bitops2.nim#L21
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set.nim#L51
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/base10.nim#L16
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_find.nim#L11
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/interval_set.nim#L111
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_rotate.nim#L14
https://github.com/status-im/nim-stew/blob/d7a6868ba84165e7fdde427af9a1fc3f5f5cc151/stew/sorted_set/rbtree_walk.nim#L11

stew/base36.nim
gmelodie: They seem to be required (base64, base58, base32 have the same functions) I'm marking them as public

tersec: None of those modules is a good example to copy.

stew/base36.nim
gmelodie: Yup that's it

stew/base36.nim
```diff
@@ -0,0 +1,222 @@
...
+    const alphabet = B36LcAlphabet
+  elif (btype is Base36Uc):
+    const alphabet = B36UcAlphabet
+
+  let binsz = inbytes.len
+  var zcount = 0
+  while zcount < binsz and inbytes[zcount] == 0x00'u8:
+    inc(zcount)
+
+  let size = ((binsz - zcount) * 277) div 179 + 1
```
tersec, line 73: This could use `encodedLength` and centralize the calculation in one place.

stew/base36.nim
```diff
@@ -0,0 +1,222 @@
...
+    inc(i)
+
+  return Base36Status.Success
+
+
+func encode*(btype: typedesc[Base36Types],
+             inbytes: openArray[byte]): string {.inline.} =
+  ## Encode array of bytes ``inbytes`` using Base36 encoding and return
+  ## encoded string.
+  var size = (len(inbytes) * 277) div 179 + 1 + 1
```
tersec, line 113: Is the additional `+ 1` over `encodedLength` a maximum value? This in any case could also use `encodedLength`, since right now basically the same calculation is written in 3 different places.

stew/base36.nim
tersec: https://github.com/status-im/nim-stew/compare/89498a6349addf3205cfefff8642f3d6f0f748c0..3c74cf03aea444e4ac462bf04ea957d4e6f3dd53

stew/base36.nim
tersec: https://github.com/status-im/nim-stew/compare/89498a6349addf3205cfefff8642f3d6f0f748c0..3c74cf03aea444e4ac462bf04ea957d4e6f3dd53

stew/base36.nim
```diff
@@ -0,0 +1,222 @@
...
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+## This module implements Base36 encoding and decoding procedures.
+## This module supports two Base36 alphabets: lowercase `Base36Lc` and uppercase `Base36Uc`.
+
+## This module follows Base36 as defined in https://github.com/multiformats/multibase/blob/f378d3427fe125057facdbac936c4215cc777920/rfcs/Base36.md
```
tersec, line 12: https://status-im.github.io/nim-style-guide/formatting.style.html#style-formattingstyle
> We strive to follow [NEP-1](https://nim-lang.org/docs/nep1.html) for style matters - naming, capitalization, 80-character limit etc.

Obviously, this doesn't mean breaking long URLs into multiple pieces; some lexical units are atomic. And in practice, we don't always use strict 80-character lines; "strive" there does point to the non-absoluteness of this guideline. But this one is a bit long without putting the URL on a separate line.

stew/base36.nim
```diff
@@ -0,0 +1,222 @@
...
+  Base36* = object
+  Base36Uc* = object
+  Base36Lc* = object
+
+  Base36Types* = Base36 | Base36Uc | Base36Lc
+
+  Base36Error* = object of CatchableError
+    ## Base36 specific exception type
+
+func newAlphabet36(s: string): Base36Alphabet =
```
tersec, line 36: It would be useful here to indicate explicitly in the type that it's an `s: static[string]`, i.e. because the intent isn't for this to be used in a random runtime way but as part of a compile-time transformation of one kind of constant to another kind of constant.

stew/base36.nim
tersec: Or because of the `k`?

tersec (approved):

PR #246, opened 2025-05-12
bump version to nim-stew v0.4.0

gmelodie:
We need a new version to refer to in `nim-libp2p`'s `.nimble file (recommended as per [@arnetheduck's comment](https://discord.com/channels/613988663034118151/740631022080622745/1371484042800664671)) in order to import `base36` functionalities.

### DIFF
--- a/stew.nimble
+++ b/stew.nimble
@@ -1,7 +1,7 @@
 mode = ScriptMode.Verbose
 
 packageName   = "stew"
-version       = "0.3.0"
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

tersec:
Ok, the next step is also to create a tag e.g., as in https://github.com/status-im/nim-stew/tags but this probably needs to happen after merging

PR #248, opened 2025-06-18
return `var openArray` from `makeOpenArray` in buggy nim versions

arnetheduck:
Workaround from https://github.com/nim-lang/Nim/issues/19171


arnetheduck:
seems like more syntax than it's worth? ie the extra temporary is harmless given it's just a simple pointer in the end.

PR #256, opened 2025-08-27
importops: add `tryImport` for testing if an import exists

arnetheduck:


tests/test_importops.nim
```diff
@@ -0,0 +1,14 @@
...
+
+when tryImport std / tables:
+  discard Table[string, string]() # avoid unused warning
+else:
+  {.error: "tables should exist".}
+
+when tryImport std / sets as s:
+  discard s.HashSet[string]()
+else:
+  {.error: "tables should exist".}
```
tersec, line 11: I guess, `sets should exist`

PR #258, opened 2025-09-01
sequtils: findIt

arnetheduck:
https://github.com/nim-lang/Nim/pull/25134

### DIFF
--- a/stew/shims/sequtils.nim
+++ b/stew/shims/sequtils.nim
@@ -1,0 +1,21 @@
+# stew
+# Copyright 2025 Status Research & Development GmbH
+# Licensed under either of
+#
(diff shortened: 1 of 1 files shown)


tersec:
Such as, just as a hypothetical example, https://github.com/nim-lang/Nim/pull/25134/commits/33796fa7f40d38e8937e94c00061047152552e9b

tersec:
That said, I actually find the required `items(...)` wrapping to be one of the annoying parts of the `std/sequtils` algorithms, because it makes composing them or using them on non-`openArray`s pointlessly harder. But yeah, it's consistent with what they do.

arnetheduck:
yeah, in the end I think both the api and implementation are actually fixed by anyIt/find and their various quirks .. I just forgot `for` isn't the one injecting the iteration index but rather it's for calling `pairs` which we can't use here.

arnetheduck:
Upstream merged

